PR #50, opened 2019-07-31
feat(notifications): Add notifications for decisions.

mikeng13:
# Summary
- Introduces notification center and manager for handling notifications internally in a consistent way
- Each top-level component can offer hooks for developers to subscribe to, as shown with the `CompositeService`

Due to the lack of generics in Go, I've opted to use the only generic construct available, which is `interface{}`. The notification center and manager don't really care about what payload gets sent back and forth. In this implementation, the component that houses the notification center will take care of type-casting the payload to a type-specific notification that will be sent back to the caller.

This is WIP as I am trying to get early feedback on it.

# Tests
- Unit tests


optimizely/decision/composite_service.go
```diff
@@ -18,38 +18,56 @@ package decision
...
+	if s.notificationCenter != nil {
+		s.notificationCenter.Send(notification.Decision, featureDecision)
+	}
+	return featureDecision, err
+}
+
+// OnDecision registers a handler for Decision notifications
+func (s CompositeService) OnDecision(callback func(notification.DecisionNotification)) {
+	handler := func(payload interface{}) {
+		if decisionNotification, ok := payload.(notification.DecisionNotification); ok {
```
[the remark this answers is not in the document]
mikeng13: Yeah we should

optimizely/decision/composite_service.go
```diff
@@ -18,38 +18,56 @@ package decision
...
-			return featureDecision, err
+			break
 		}
 	}
 
 	// @TODO: add errors
-	return featureDecision, nil
+
+	// @TODO convert the decision to a notification.DecisionNotification
+	if s.notificationCenter != nil {
```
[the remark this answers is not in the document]
mikeng13: Hmm good question. We don't always control the instantiation of the implementation. That is, a user can always just say `s := decision.CompositeService{}` and pass that into the client and not pass in a notification center instead of calling our `decision.CompositeService()` factory method. The question is then, how well do we guard against these possibilities since?

optimizely/notification/manager.go
```diff
@@ -0,0 +1,47 @@
...
+
+// AddHandler adds the given handler
+func (am *AtomicManager) AddHandler(newHandler func(interface{})) (int, error) {
+	atomic.AddUint32(&am.counter, 1)
+	am.handlers[am.counter] = newHandler
+	return int(am.counter), nil
+}
+
+// Send sends the notification to the registered handlers
+func (am *AtomicManager) Send(notification interface{}) {
```
aliabbasrizvi: Curious. Is there a reason why `Send` is highlighted differently in Github? Is it a keyword by any chance?

optimizely/notification/entities.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Type is the type of notification
+type Type string
+
+const (
+	// Decision notification type
+	Decision Type = "decision"
+)
+
+// DecisionNotification is a notification triggered when a decision is made
```
aliabbasrizvi: nit. I think we can do a better job documenting `DecisionNotification`

optimizely/notification/center.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Send sends the given notification payload to all listeners of type
+func (c *DefaultCenter) Send(notificationType Type, notification interface{}) error {
+	if manager, ok := c.managerMap[notificationType]; ok {
+		manager.Send(notification)
+		return nil
+	}
+
+	return fmt.Errorf("No notification manager found for type %s", notificationType)
```
aliabbasrizvi: Lets be consistent in terminology and say notification handler and not manager.

optimizely/notification/center.go
```diff
@@ -0,0 +1,59 @@
...
+	if manager, ok := c.managerMap[notificationType]; ok {
+		return manager.AddHandler(handler)
+	}
+
+	return -1, fmt.Errorf("No notification manager found for type %s", notificationType)
+}
+
+// Send sends the given notification payload to all listeners of type
+func (c *DefaultCenter) Send(notificationType Type, notification interface{}) error {
+	if manager, ok := c.managerMap[notificationType]; ok {
```
aliabbasrizvi: We should call this handler and not manager right?

optimizely/notification/manager.go
mikeng13: It's not a keyword, not sure why it's highlighted differently

optimizely/notification/center.go
mikeng13: Hmm no, this should be consistent with our Java implementation, where we have a manager per notification type that manages all the handlers for that type.

optimizely/notification/center.go
mikeng13: Same as above

optimizely/notification/entities.go
mikeng13: Will do.

optimizely/decision/composite_service.go
```diff
@@ -17,39 +17,74 @@
...
 	}
 }
 
 // GetFeatureDecision returns a decision for the given feature key
 func (s CompositeService) GetFeatureDecision(featureDecisionContext FeatureDecisionContext, userContext entities.UserContext) (FeatureDecision, error) {
 	var featureDecision FeatureDecision
-
+	var err error
 	// loop through the different features decision services until we get a decision
 	for _, decisionService := range s.featureDecisionServices {
```
pawels-optimizely: strange flow...

optimizely/decision/composite_service.go
```diff
@@ -18,38 +18,56 @@ package decision
...
+
+	// @TODO convert the decision to a notification.DecisionNotification
+	if s.notificationCenter != nil {
+		s.notificationCenter.Send(notification.Decision, featureDecision)
+	}
+	return featureDecision, err
+}
+
+// OnDecision registers a handler for Decision notifications
+func (s CompositeService) OnDecision(callback func(notification.DecisionNotification)) {
```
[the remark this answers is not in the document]
mikeng13: Synced offline about making notification center / manager an internal component while each top-level component exposes what it notifies on.

optimizely/decision/composite_service.go
```diff
@@ -17,39 +17,74 @@
...
-			return featureDecision, err
+			break
 		}
 	}
 
 	// @TODO: add errors
-	return featureDecision, nil
+	if s.notificationCenter != nil {
+		decisionInfo := map[string]interface{}{
+			"feature": map[string]interface{}{
```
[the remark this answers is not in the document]
pawels-optimizely: if all the fields are known and set then probably yes.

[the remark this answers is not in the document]
mikeng13: The trick here is that the decisionInfo, as designed, can have arbitrary keys depending on whether it came from an activate or feature accessor

PR #51, opened 2019-08-01
feat(err): Panic handling in the IsFeatureEnabled API.

mikeng13:
# Summary
- Introduces a top-level defer guard to ensure we recover from any panicking when using the IsFeatureEnabled API

# Tests
- Unit tests

### DIFF
--- a/optimizely/client/client.go
+++ b/optimizely/client/client.go
@@ -19,6 +19,7 @@ package client
 import (
 	"errors"
 	"fmt"
+	"runtime/debug"
 
 	"github.com/optimizely/go-sdk/optimizely"
 	"github.com/optimizely/go-sdk/optimizely/decision"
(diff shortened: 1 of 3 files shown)


optimizely/client/client.go
```diff
@@ -36,19 +37,28 @@ type OptimizelyClient struct {
...
 		logger.Error(errorMessage, nil)
 		return false, err
 	}
 
+	defer func() {
+		if r := recover(); r != nil {
+			errorMessage := fmt.Sprintf(`Optimizely SDK is panicking with the error "%s"`, string(debug.Stack()))
+			err = errors.New(errorMessage)
+			logger.Error(errorMessage, err)
+			result = false
```
pawels-optimizely: why are you assigning result to false? what is the purpose of it?

mikeng13: You are right, this is actually not needed because the zero-value is `false`

PR #52, opened 2019-08-02
feat: Add simple client factory method and remove metrics logic

pawels-optimizely:


optimizely/client/factory.go
```diff
@@ -85,3 +87,28 @@ func (f OptimizelyFactory) ClientWithContext(ctx context.Context) (*OptimizelyCl
...
+// Client returns a client initialized with the defaults
+func (f OptimizelyFactory) Client() (*OptimizelyClient, error) {
+	var configManager optimizely.ProjectConfigManager
+
+	if f.SDKKey != "" {
+		url := fmt.Sprintf(CDNTemplate, f.SDKKey)
+		request := config.NewRequester(url)
+		ctx := context.Background()
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithCancel(ctx)
```
mikeng13: Should we pass the `cancelFunc` to the client in the other factory method where the user passes in context? They should be able to call `optimizelyClient.Stop()` with the client created `withContext` as well

optimizely/client/factory.go
```diff
@@ -25,18 +25,21 @@ import (
...
 
+const datafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json"
+
 // OptimizelyFactory is used to construct an instance of the OptimizelyClient
 type OptimizelyFactory struct {
-	SDKKey   string
-	Datafile []byte
+	SDKKey               string
+	Datafile             []byte
+	ProjectConfigManager optimizely.ProjectConfigManager
```
mjc1283: In Java the factory has only static methods, and the only component shared among instances it creates is the logger. Are we sure we want to do this instead (associating a project config manager with the factory and applying it to all instances created in factory methods)?

mikeng13: Hmm good point. I will go get some inspiration and refactor this.

mjc1283: I like the changes, but is `OptimizelyFactory` struct still supposed to have this `ProjectConfigManager` field?

optimizely/client/factory.go
```diff
@@ -18,25 +18,34 @@ package client
...
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/optimizely/go-sdk/optimizely"
 	"github.com/optimizely/go-sdk/optimizely/config"
 	"github.com/optimizely/go-sdk/optimizely/decision"
 )
 
+const datafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json"
```
mikeng13: I have code stashed that moves this out of the client factory. Will put up in a separate PR.

PR #53, opened 2019-08-05
feat(coveralls): coveralls and go-report-card status.

yasirfolio3:


mikeng13:
The coveralls link doesn't show anything though

yasirfolio3:
> The coveralls doesn't show anything, should it?

badge should work once coveralls receives a build. Since its on a private repo as of now, we need COVERALLS_TOKEN to integrate it! more info given here:
https://github.com/mattn/goveralls

yasirfolio3:
> The coveralls doesn't show anything, should it?

it will after merging into go-alpha as can be seen here:
https://coveralls.io/github/optimizely/go-sdk

PR #56, opened 2019-08-08
(chore): add batch event sending

thomaszurkan-optimizely:
This adds event batch.  The processor saves user events to its queue until flush time, at that time it attempts to get the batch size of events (default 20) from the queue, it attempts to batch them.  If a batch fails, it just sends whatever was already in that batch and deletes that amount from the queue.  Then, it gets the next lot and attempts to batch and send again until the queue is empty.


optimizely/event/processor.go
```diff
@@ -24,6 +24,7 @@ type QueueingEventProcessor struct {
 
 func NewEventProcessor(queueSize int, flushInterval time.Duration ) Processor {
 	p := &QueueingEventProcessor{MaxQueueSize: queueSize, FlushInterval:flushInterval, Q:NewInMemoryQueue(queueSize), EventDispatcher:&HttpEventDispatcher{}}
+	p.BatchSize = 20
```
mikeng13: Default batch size is 10 in other SDKs. We should prob keep it consistent

optimizely/event/processor.go
```diff
@@ -63,35 +64,69 @@ func (p *QueueingEventProcessor) StartTicker() {
 	}()
 }
 
+func (p *QueueingEventProcessor) Append(current *Batch, new Batch) bool {
+	if current.ProjectID == new.ProjectID &&
+		current.Revision == new.Revision {
+		visitors := append(current.Visitors, new.Visitors[0])
```
mikeng13: How come we need to create a new batch object just to hold 1 visitor? Can we just accept a visitor as argument and append it?

optimizely/event/processor.go
```diff
@@ -63,35 +64,69 @@ func (p *QueueingEventProcessor) StartTicker() {
...
 	// however, if there is a ticker cycle already processing, we should wait
 	p.Mux.Lock()
+	var batchEvent Batch
+	var batchEventCount = 0
+	var failedToSend = false
+
 	for p.EventsCount() > 0 {
-		events := p.GetEvents(1)
+		if failedToSend {
+			break
```
mikeng13: What does this mean? Should we be logging a message or something?

optimizely/event/processor.go
```diff
@@ -63,35 +64,69 @@ func (p *QueueingEventProcessor) StartTicker() {
 	}()
 }
 
+func (p *QueueingEventProcessor) Append(current *Batch, new Batch) bool {
```
mikeng13: add doc string on what this does

optimizely/event/processor.go
```diff
@@ -63,35 +64,69 @@ func (p *QueueingEventProcessor) StartTicker() {
...
+					if batchEventCount >= p.BatchSize {
+						break
+					}
 				}
 			}
 		}
- 	}
+		if batchEventCount > 0 {
+			p.EventDispatcher.DispatchEvent(createLogEvent(batchEvent), func(success bool) {
+				fmt.Println(success)
```
mikeng13: Should probably remove this `Println`

optimizely/event/processor.go
```diff
@@ -63,35 +64,69 @@ func (p *QueueingEventProcessor) StartTicker() {
...
+
+					if batchEventCount >= p.BatchSize {
+						break
+					}
 				}
 			}
 		}
- 	}
+		if batchEventCount > 0 {
+			p.EventDispatcher.DispatchEvent(createLogEvent(batchEvent), func(success bool) {
```
mikeng13: Should this blocking operation happen asynchronously via a separate go routine? Or does `FlushEvents` already happen in another Go routine?

optimizely/event/processor.go
```diff
@@ -63,35 +64,69 @@ func (p *QueueingEventProcessor) StartTicker() {
...
-						fmt.Println(success)
-						if success {
-							p.Remove(1)
-						}
-					})
+			for i := 0; i < len(events); i++ {
+				userEvent, ok := events[i].(UserEvent)
+				if ok {
+					var eventToAdd Batch
+					if userEvent.Conversion != nil {
```
mikeng13: I think the logic here can be rearranged a little so we create a visitor object first and only create the batch event if `batchEventCount == 0`

optimizely/event/events.go
```diff
@@ -1,6 +1,6 @@
 package event
 
-type EventContext struct {
+type Context struct {
```
mikeng13: Ah, linting must've kicked in for you 😄

optimizely/event/processor.go
thomaszurkan-optimizely: the flush is called in a go routine.

optimizely/event/factory.go
```diff
@@ -67,15 +67,15 @@ func CreateImpressionUserEvent(context EventContext, experiment entities.Experim
 	userEvent := UserEvent{}
 	userEvent.Timestamp = makeTimestamp()
 	userEvent.VisitorID = userContext.ID
-	userEvent.UUID = guuid.New().String()
+	userEvent.Uuid = guuid.New().String()
```
mikeng13: nit: This is a little inconsistent with the `EventBatch.UUID`. I think the Go linter complains about `Uuid` having to be `UUID` since it is an acronym

optimizely/event/processor.go
```diff
@@ -68,35 +72,73 @@ func (p *QueueingEventProcessor) StartTicker() {
 	}()
 }
 
+// check if user event can be batched in the current batch
+func (p *QueueingEventProcessor)canBatch(current *Batch, user UserEvent) bool {
```
mikeng13: nit: there needs to be a space before `canBatch`

optimizely/event/processor.go
```diff
@@ -68,35 +72,73 @@ func (p *QueueingEventProcessor) StartTicker() {
...
+	if current.ProjectID == user.EventContext.ProjectID &&
+		current.Revision == user.EventContext.Revision {
+		return true
+	}
+
+	return false
+}
+
+// add the visitor to the current batch
+func (p *QueueingEventProcessor)addToBatch(current *Batch, visitor Visitor) {
```
mikeng13: nit: ditto above

optimizely/event/processor_test.go
```diff
@@ -30,27 +30,103 @@ func TestDefaultEventProcessor_ProcessImpression(t *testing.T) {
 
 }
 
-func TestDefaultEventProcessor_ProcessImpressions(t *testing.T) {
-	processor := NewEventProcessor(100, 100)
+type FakeDispatcher struct {
```
mikeng13: nit: I've been calling everything mock*. So this can be MockDispatcher. 

Also, I've been using the `mock` function from the `testify` package to assert that function calls are made.

PR #60, opened 2019-08-13
feat(log): Add loglevel to default logger message

YuanC:
## Summary
- Modified the `log` function in `logger.go` to prepend the log level before the message

## Test plan
Modified unit tests in `logger_test.go` to include log level

## Issues
[OASIS-5171](https://optimizely.atlassian.net/browse/OASIS-5171)


optimizely/logging/logger.go
```diff
@@ -70,7 +70,14 @@ func (p NamedLogProducer) Error(message string, err interface{}) {
 }
 
 func (p NamedLogProducer) log(logLevel int, message string) {
-	// prepends the name to the message
-	message = fmt.Sprintf("[%s] %s", p.name, message)
+	logLevelStrings := map[int]string{
```
mikeng13: There is a cooler approach to this since we are using enums: https://programming.guide/go/define-enumeration-string.html

😄

PR #61, opened 2019-08-13
refac(config-manager): simplifies the creation of polling config manager with options

mikeng13:
# Summary
- Allows creation of polling config manager with defaults instead of having to always pass in a requester object

# Tests
- Unit tests

### DIFF
--- a/optimizely/config/polling_manager.go
+++ b/optimizely/config/polling_manager.go
@@ -36,9 +36,9 @@ var cmLogger = logging.GetLogger("PollingConfigManager")
 
 // PollingProjectConfigManagerOptions used to create an instance with custom configuration
 type PollingProjectConfigManagerOptions struct {
-	FallbackDatafile []byte
-	PollingInterval  time.Duration
-	Requester        Requester
+	Datafile        []byte
+	PollingInterval time.Duration
+	Requester       Requester
 }
 
 // PollingProjectConfigManager maintains a dynamic copy of the project config
@@ -111,7 +111,7 @@ func NewPollingProjectConfigManagerWithOptions(ctx context.Context, sdkKey strin
 
 	pollingProjectConfigManager := PollingProjectConfigManager{requester: requester, pollingInterval: pollingInterval, ctx: ctx}
 
-	pollingProjectConfigManager.activate(options.FallbackDatafile, true) // initial poll
+	pollingProjectConfigManager.activate(options.Datafile, true) // initial poll
 
 	cmLogger.Debug("Polling Config Manager Initiated")
 	go pollingProjectConfigManager.activate([]byte{}, false)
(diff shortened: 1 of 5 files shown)


optimizely/config/polling_manager.go
```diff
@@ -83,22 +93,38 @@ func (cm *PollingProjectConfigManager) activate(initialPayload []byte, init bool
...
+// NewPollingProjectConfigManagerWithOptions returns new instance of PollingProjectConfigManager with the given options
+func NewPollingProjectConfigManagerWithOptions(ctx context.Context, sdkKey string, options PollingProjectConfigManagerOptions) *PollingProjectConfigManager {
 
-	if pollingWait == 0 {
-		pollingWait = defaultPollingWait
+	var requester Requester
+	if options.Requester != nil {
+		requester = options.Requester
+	} else {
+		url := fmt.Sprintf(DatafileURLTemplate, sdkKey)
```
pawels-optimizely: are we sure at this point sdkKey cannot be empty?

optimizely/config/polling_manager.go
```diff
@@ -27,16 +27,26 @@ import (
...
+const defaultPollingInterval = time.Duration(5 * time.Minute) // default to 5 minutes for polling
+
+// DatafileURLTemplate is used to construct the endpoint for retrieving the datafile from the CDN
+const DatafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json"
 
 var cmLogger = logging.GetLogger("PollingConfigManager")
 
+// PollingProjectConfigManagerOptions used to create an instance with custom configuration
+type PollingProjectConfigManagerOptions struct {
+	FallbackDatafile []byte
```
aliabbasrizvi: Why not just call this `Datafile`? It is technically not the fallback datafile but the initial datafile so if not `Datafile` may be `InitialDatafile` will be more accurate than `FallbackDatafile` because that suggests that if network is not reachable then that is the datafile it will fall back to which will not be the case it will be datafile that was most recently polled. Does that make sense?

optimizely/config/polling_manager.go
mikeng13: discussed offline, the check for the SDK key happens upstream so we're comfortable not doing it here

optimizely/config/polling_manager.go
```diff
@@ -83,22 +93,38 @@ func (cm *PollingProjectConfigManager) activate(initialPayload []byte, init bool
...
+	if options.Requester != nil {
+		requester = options.Requester
+	} else {
+		url := fmt.Sprintf(DatafileURLTemplate, sdkKey)
+		requester = NewHTTPRequester(url)
 	}
 
-	pollingProjectConfigManager := PollingProjectConfigManager{requester: requester, pollingWait: pollingWait, ctx: ctx}
+	pollingInterval := options.PollingInterval
+	if pollingInterval == 0 {
```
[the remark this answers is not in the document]
mikeng13: Yeah. I'll do in a follow-up PR.

optimizely/config/polling_manager.go
mikeng13: Yeah I'll make that change

PR #63, opened 2019-08-15
feat(getenabledfeatures): Add GetEnabledFeatures API

YuanC:
# Summary
- Add `GetEnabledFeatures` as a function in `client.go`
- Add `GetFeatureList()` function in `datafileprojectconfig/config.go`

# Tests
- Unit tests in `client_test.go`
  - `TestGetEnabledFeatures` checks that the function returns 1 out of 2 features
  - `TestGetEnabledFeaturesErrorCases` checks for error upon invalid Optimizely instance
  - `TestGetEnabledFeaturesPanic` tests for panic and recovery in the function

## Issues
[OASIS-5107](https://optimizely.atlassian.net/browse/OASIS-5107)

### DIFF
--- a/optimizely/client/client.go
+++ b/optimizely/client/client.go
@@ -125,7 +125,6 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext) 
 		}
 	}
 
-	// @TODO(mng): send impression event
 	return enabledFeatures, nil
 }
 
--- a/optimizely/client/client.go
+++ b/optimizely/client/client.go
@@ -93,6 +93,42 @@ func (o *OptimizelyClient) IsFeatureEnabled(featureKey string, userContext entit
 	return result, nil
 }
 
+// GetEnabledFeatures returns an array containing the keys of all features in the project that are enabled for the given user.
+func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext) (enabledFeatures []string, err error) {
+	if !o.isValid {
+		errorMessage := "Optimizely instance is not valid. Failing GetEnabledFeatures."
+		err := errors.New(errorMessage)
+		logger.Error(errorMessage, nil)
+		return enabledFeatures, err
+	}
+
+	defer func() {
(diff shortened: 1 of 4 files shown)


optimizely/interface.go
```diff
@@ -31,6 +31,7 @@ type ProjectConfig interface {
 	GetEventByKey(string) (entities.Event, error)
 	GetExperimentByKey(string) (entities.Experiment, error)
 	GetFeatureByKey(string) (entities.Feature, error)
+	GetFeatureMap() map[string]entities.Feature
```
mikeng13: Hmm, this is a proposed public API. I think what is more relevant is that we get the list of all features, rather than just the map. So let's expose a `GetFeatureList` API instead and use that in the client. @mjc1283 FYI

optimizely/client/client.go
```diff
@@ -93,6 +93,42 @@ func (o *OptimizelyClient) IsFeatureEnabled(featureKey string, userContext entit
...
+	}()
+
+	projectConfig := o.configManager.GetConfig()
+
+	if reflect.ValueOf(projectConfig).IsNil() {
+		return enabledFeatures, fmt.Errorf("project config is null")
+	}
+
+	features := projectConfig.GetFeatureMap()
+	for feature := range features {
```
mikeng13: To be more precise, it is a `featureKey` and not a `feature`

optimizely/client/client.go
```diff
@@ -93,6 +93,42 @@ func (o *OptimizelyClient) IsFeatureEnabled(featureKey string, userContext entit
...
+	featureList := projectConfig.GetFeatureList()
+	for _, feature := range featureList {
+		isEnabled, _ := o.IsFeatureEnabled(feature.Key, userContext)
+
+		if isEnabled {
+			enabledFeatures = append(enabledFeatures, feature.Key)
+		}
+	}
+
+	// @TODO(mng): send impression event
```
mikeng13: Nit: this comment doesn't apply here

PR #65, opened 2019-08-16
feat(config): Add missing config properties for events.

mikeng13:
# Summary
- This is the first part of tying the event processor to the client. 
- Adds client and version name to the config
- Parses attributes out of the datafile to be used to build the event payload
- Refactors the event factory to use the project config since it is now a top-level component

# Tests
- Unit tests


optimizely/config/datafileprojectconfig/config.go
```diff
@@ -82,11 +87,16 @@ func NewDatafileProjectConfig(jsonDatafile []byte) (*DatafileProjectConfig, erro
...
 
+	attributeMap, attributeKeyToIDMap := mappers.MapAttributes(datafile.Attributes)
 	experiments, experimentKeyMap := mappers.MapExperiments(datafile.Experiments)
 	rolloutMap := mappers.MapRollouts(datafile.Rollouts)
 	mergedAudiences := append(datafile.TypedAudiences, datafile.Audiences...)
 	config := &DatafileProjectConfig{
 		audienceMap:          mappers.MapAudiences(mergedAudiences),
+		attributeMap:         attributeMap,
+		attributeKeyToIDMap:  attributeKeyToIDMap,
+		clientName:           optimizely.ClientName,
```
mjc1283: `clientName` isn't in the datafile - should it be part of the instance or `Optimizely` instead?

optimizely/event/factory.go
```diff
@@ -200,34 +200,34 @@ func createBatchEvent(userEvent UserEvent, visitor Visitor) Batch {
...
-			attribute.EntityID = id
+		visitorAttribute := VisitorAttribute{}
+		attribute, _ := projectConfig.GetAttributeByKey(key)
+		if attribute.ID != "" {
+			visitorAttribute.EntityID = attribute.ID
 		} else if strings.HasPrefix(key, specialPrefix) {
-			attribute.EntityID = key
+			visitorAttribute.EntityID = key
 		} else {
 			continue
```
mjc1283: If we reach this point does it mean the user passed in an invalid attribute? Should we log? IIRC in other SDKs we validate at a higher level, so would never expect this to happen when building an event.

optimizely/config/datafileprojectconfig/config.go
mikeng13: Ah yeah good point. I'll move it out and maybe pass it to the event factory since that is where it is needed

optimizely/event/factory.go
mikeng13: Good point. I'll see if we can take care of it before this. Otherwise we'll log here

mikeng13: The filtering actually happens when building the event: https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/core/event_builder/index.js#L64 - which makes sense because we want to preserve the attributes when we send it back via the notification listener (for activate or track)

PR #66, opened 2019-08-16
fix(testapp-support): Fixes to allow test-app to use notification listeners.

yasirfolio3:
# Summary
- Added DecisionService as an option to OptimizelyClient custom configuration.
- Fixed casing for Source Rollout.


optimizely/client/factory.go
```diff
@@ -105,7 +106,12 @@ func (f OptimizelyFactory) ClientWithOptions(clientOptions Options) (*Optimizely
 	}
 
 	// @TODO: allow decision service to be passed in via options
```
mikeng13: Nit: remove this comment

yasirfolio3: done.

PR #67, opened 2019-08-19
feat: Hook up event processor to IsFeatureEnabled.

mikeng13:
# Summary

This PR enables sending impressions for features tests from the `IsFeatureEnabled` API: 

- Adds experiments to the feature model in the project config
- Re-enabled the experiment decision service in the composite feature service to evaluate feature tests
- Integrate the event processor with the Optimizely client and send event to it from the IsFeatureEnabled API

# Tests
- Unit

### DIFF
--- a/optimizely/client/factory.go
+++ b/optimizely/client/factory.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/optimizely/go-sdk/optimizely/event"
 
@@ -44,7 +45,7 @@ type OptimizelyFactory struct {
 }
 
 const defaultEventQueueSize = 10
-const defaultEventFlushInterval = 30000
+const defaultEventFlushInterval = 30 * time.Second
 
 // StaticClient returns a client initialized with a static project config
 func (f OptimizelyFactory) StaticClient() (*OptimizelyClient, error) {
--- a/optimizely/client/factory.go
+++ b/optimizely/client/factory.go
@@ -116,6 +116,8 @@ func (f OptimizelyFactory) ClientWithOptions(clientOptions Options) (*Optimizely
 		client.decisionService = decision.NewCompositeService(notificationCenter)
 	}
 
+	// @TODO: allow event processor to be passed in
+	// @TODO: pass the context object to the event processor
 	client.eventProcessor = event.NewEventProcessor(defaultEventQueueSize, defaultEventFlushInterval)
 	client.isValid = true
 	return client, nil
--- a/optimizely/client/factory.go
+++ b/optimizely/client/factory.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/optimizely/go-sdk/optimizely/event"
+
 	"github.com/optimizely/go-sdk/optimizely/notification"
 
 	"github.com/optimizely/go-sdk/optimizely"
@@ -40,6 +42,9 @@ type OptimizelyFactory struct {
 	SDKKey   string
 	Datafile []byte
 }
+
+const defaultEventQueueSize = 10
+const defaultEventFlushInterval = 30000
 
 // StaticClient returns a client initialized with a static project config
 func (f OptimizelyFactory) StaticClient() (*OptimizelyClient, error) {
@@ -111,6 +116,7 @@ func (f OptimizelyFactory) ClientWithOptions(clientOptions Options) (*Optimizely
 		client.decisionService = decision.NewCompositeService(notificationCenter)
 	}
 
+	client.eventProcessor = event.NewEventProcessor(defaultEventQueueSize, defaultEventFlushInterval)
 	client.isValid = true
 	return client, nil
 }
--- a/optimizely/decision/composite_feature_service.go
+++ b/optimizely/decision/composite_feature_service.go
@@ -23,7 +23,7 @@ import (
 	"github.com/optimizely/go-sdk/optimizely/logging"
 )
 
-var cfLogger = logging.GetLogger("ComposuteFeatureService")
+var cfLogger = logging.GetLogger("CompositeFeatureService")
 
 // CompositeFeatureService is the default out-of-the-box feature decision service
 type CompositeFeatureService struct {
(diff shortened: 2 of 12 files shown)


optimizely/client/factory.go
```diff
@@ -111,6 +116,7 @@ func (f OptimizelyFactory) ClientWithOptions(clientOptions Options) (*Optimizely
 		client.decisionService = decision.NewCompositeService(notificationCenter)
 	}
 
+	client.eventProcessor = event.NewEventProcessor(defaultEventQueueSize, defaultEventFlushInterval)
```
mjc1283: - Are we going to add more configurability, like allowing passing in event processor, or event processor options?
- Should `ctx` be provided to the event processor, so that `cancelFunc` can shut it down also?

optimizely/decision/composite_feature_service.go
```diff
@@ -47,12 +53,24 @@ func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionCont
 		}
 
 		experimentDecision, err := f.experimentDecisionService.GetDecision(experimentDecisionContext, userContext)
-		featureDecision := FeatureDecision{
-			Experiment: experiment,
-			Decision:   experimentDecision.Decision,
-			Variation:  experimentDecision.Variation,
+		// only return the decision if we have a valid variation
+		// @TODO: log the decision made here as debug
+		if experimentDecision.Variation.ID != "" {
```
mjc1283: Nit: is it weird that `ID` could be an empty string? This seems like invalid data. How does this end up being an empty string and is there a better representation we can use?

optimizely/decision/composite_feature_service.go
```diff
@@ -47,12 +53,24 @@ func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionCont
 		}
 
 		experimentDecision, err := f.experimentDecisionService.GetDecision(experimentDecisionContext, userContext)
-		featureDecision := FeatureDecision{
-			Experiment: experiment,
-			Decision:   experimentDecision.Decision,
-			Variation:  experimentDecision.Variation,
+		// only return the decision if we have a valid variation
+		// @TODO: log the decision made here as debug
```
mjc1283: TODO seems already done?

optimizely/event/processor.go
```diff
@@ -26,8 +27,8 @@ type QueueingEventProcessor struct {
 var pLogger = logging.GetLogger("EventProcessor")
 
 // NewEventProcessor returns a new instance of QueueingEventProcessor with queueSize and flushInterval
-func NewEventProcessor(queueSize int, flushInterval time.Duration ) Processor {
-	p := &QueueingEventProcessor{MaxQueueSize: queueSize, FlushInterval:flushInterval, Q:NewInMemoryQueue(queueSize), EventDispatcher:&HTTPEventDispatcher{}}
+func NewEventProcessor(queueSize int, flushInterval time.Duration) *QueueingEventProcessor {
```
mjc1283: Since we're now using the event processor, should we make it behave like the other SDKs, or alternatively update other SDKs to be like this one?

Unless I misunderstand, this event processor is unique because 1) it doesn't reset the buffer until the dispatcher calls the callback with `true`, and 2) it retries sending the batch again if the dispatcher calls the callback with `false`.

IMO, we should remove this retrying and leave it to the event dispatcher.

cc @aliabbasrizvi

optimizely/client/factory.go
mikeng13: Yes to both. I am planning to add more configurability later on as people start asking about it during beta. 

The ctx should also be passed in but that requires a small refactor of event processor that we can take up in a separate PR (I think Tom might be looking at taking in the context already as part of his work).

Regardless, I will add the TODOs here so we know to go back and add em.

optimizely/decision/composite_feature_service.go
mikeng13: Empty string is the zero-value for strings and it is what the variation is initialized with. I toyed with using `nil` for the variation but for that we would need to pass references to the Variation instead of the copy of the Variation because of how it works in Go. Making the function return the reference to the Variation would make it seem like we are returning a mutable struct, which we are not.

optimizely/decision/composite_feature_service.go
mikeng13: yeah I'll remove that

optimizely/event/processor.go
mikeng13: 👍 I agree. Will sync with Tom on fixing this as part of his remaining work with the event processor.

optimizely/client/factory.go
```diff
@@ -41,6 +43,9 @@ type OptimizelyFactory struct {
 	Datafile []byte
 }
 
+const defaultEventQueueSize = 10
+const defaultEventFlushInterval = 30000
```
aliabbasrizvi: nit. Mention unit of time.

optimizely/decision/composite_feature_service.go
```diff
@@ -17,9 +17,14 @@
 package decision
 
 import (
+	"fmt"
+
 	"github.com/optimizely/go-sdk/optimizely/entities"
+	"github.com/optimizely/go-sdk/optimizely/logging"
 )
 
+var cfLogger = logging.GetLogger("ComposuteFeatureService")
```
aliabbasrizvi: typo. Composite.

optimizely/decision/composite_feature_service.go
aliabbasrizvi: Shouldn't we confirm that there is no `err` before doing this check?

optimizely/event/dispatcher.go
```diff
@@ -23,7 +24,7 @@ var dispatcherLogger = logging.GetLogger("EventDispatcher")
 func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(success bool)) {
 
 	jsonValue, _ := json.Marshal(event.event)
-	resp, err := http.Post( event.endPoint, "application/json", bytes.NewBuffer(jsonValue))
+	resp, err := http.Post(event.endPoint, "application/json", bytes.NewBuffer(jsonValue))
```
aliabbasrizvi: I'd recommend making `"application/json"` a variable called content type or something similar.

PR #71, opened 2019-08-22
feat(getFeatureVariableString): Add getFeatureVariableString API

yasirfolio3:
# Summary
- Add `GetFeatureVariableString` as a function in `client.go`
- Add `GetFeatureFlagByKey()` function in `datafileprojectconfig/config.go`
- Add `GetVariable()` function in `FeatureFlag` entity
- Add `MapFeatureFlags()` function in `datafileprojectconfig/mappers/feature.go`

# Tests
- Unit tests in `client_test.go`

## Issues
[OASIS-5006](https://optimizely.atlassian.net/browse/OASIS-5006)

### DIFF
--- a/optimizely/client/client.go
+++ b/optimizely/client/client.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"reflect"
 	"runtime/debug"
+	"strconv"
 
 	"github.com/optimizely/go-sdk/optimizely/event"
 
@@ -78,6 +79,7 @@ func (o *OptimizelyClient) IsFeatureEnabled(featureKey string, userContext entit
 	userID := userContext.ID
 	logger.Debug(fmt.Sprintf(`Evaluating feature "%s" for user "%s".`, featureKey, userID))
 	featureDecision, err := o.decisionService.GetFeatureDecision(featureDecisionContext, userContext)
+
 	if err != nil {
 		logger.Error("Received an error while computing feature decision", err)
 		return result, err
@@ -131,6 +133,115 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext) 
 	}
 
 	return enabledFeatures, nil
+}
+
+// GetFeatureVariableString returns string feature variable value
+func (o *OptimizelyClient) GetFeatureVariableString(featureKey string, variableKey string, userContext entities.UserContext) (value string, err error) {
+	val, err := o.getFeatureVariable("", featureKey, variableKey, userContext)
+	if err == nil {
+		switch val.(type) {
+		case string:
+			return val.(string), err
+		default:
+			break
+		}
+	}
+	return "", err
+}
+
+func (o *OptimizelyClient) getFeatureVariable(valueType interface{}, featureKey string, variableKey string, userContext entities.UserContext) (value interface{}, err error) {
+	if !o.isValid {
+		errorMessage := "Optimizely instance is not valid. Failing getFeatureVariable."
+		err := errors.New(errorMessage)
+		logger.Error(errorMessage, nil)
+		return nil, err
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
+			errorMessage := fmt.Sprintf(`Optimizely SDK is panicking with the error "%s"`, string(debug.Stack()))
+			err = errors.New(errorMessage)
+			logger.Error(errorMessage, err)
+		}
+	}()
+
+	projectConfig := o.configManager.GetConfig()
+
+	if reflect.ValueOf(projectConfig).IsNil() {
+		return nil, fmt.Errorf("project config is null")
+	}
+
+	featureFlag, err := projectConfig.GetFeatureFlagByKey(featureKey)
+	if err != nil {
+		logger.Error("Error retrieving feature flag", err)
+		return nil, err
+	}
+	variable, err := featureFlag.GetVariable(variableKey)
+	if err != nil {
+		logger.Error("Error retrieving variable", err)
+		return nil, err
+	}
+
+	var featureValue = variable.DefaultValue
+
+	feature, err := projectConfig.GetFeatureByKey(featureKey)
+	if err != nil {
+		logger.Error("Error retrieving feature", err)
+		return nil, err
+	}
+	featureDecisionContext := decision.FeatureDecisionContext{
+		Feature:       &feature,
+		ProjectConfig: projectConfig,
+	}
+
+	featureDecision, err := o.decisionService.GetFeatureDecision(featureDecisionContext, userContext)
+	if err == nil {
+		for _, v := range featureDecision.Variation.Variables {
+			if v.ID == variable.ID && featureDecision.Variation.FeatureEnabled {
+				featureValue = v.Value
+				break
+			}
+		}
+	}
+
+	var typeName = ""
+	var valueParsed interface{}
+	switch valueType.(type) {
+	case string:
+		typeName = "string"
+		valueParsed = featureValue
+		break
+	case int:
+		typeName = "integer"
+		convertedValue, err := strconv.Atoi(featureValue)
+		if err == nil {
+			valueParsed = convertedValue
+		}
+		break
+	case float64:
+		typeName = "double"
+		convertedValue, err := strconv.ParseFloat(featureValue, 64)
+		if err == nil {
+			valueParsed = convertedValue
+		}
+		break
+	case bool:
+		typeName = "boolean"
+		convertedValue, err := strconv.ParseBool(featureValue)
+		if err == nil {
+			valueParsed = convertedValue
+		}
+		break
+	default:
+		break
+	}
+
+	if valueParsed == nil || variable.Type != typeName {
+		return nil, fmt.Errorf("Variable value for key %s is invalid or wrong type", variableKey)
+	}
+
+	// @TODO(yasir): send decision notification
+	return valueParsed, nil
 }
 
 // Close closes the Optimizely instance and stops any ongoing tasks from its children components
--- a/optimizely/client/client_test.go
+++ b/optimizely/client/client_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/optimizely/go-sdk/optimizely"
+	datafileEntities "github.com/optimizely/go-sdk/optimizely/config/datafileprojectconfig/entities"
 	"github.com/optimizely/go-sdk/optimizely/decision"
 	"github.com/optimizely/go-sdk/optimizely/entities"
 	"github.com/stretchr/testify/mock"
@@ -41,6 +42,11 @@ func (c *MockProjectConfig) GetFeatureByKey(featureKey string) (entities.Feature
 func (c *MockProjectConfig) GetFeatureList() []entities.Feature {
 	args := c.Called()
 	return args.Get(0).([]entities.Feature)
+}
+
+func (c *MockProjectConfig) GetFeatureFlagByKey(featureKey string) (datafileEntities.FeatureFlag, error) {
+	args := c.Called(featureKey)
+	return args.Get(0).(datafileEntities.FeatureFlag), args.Error(1)
 }
 
 type MockProjectConfigManager struct {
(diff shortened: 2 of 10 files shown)


optimizely/client/client.go
```diff
@@ -133,6 +135,115 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
...
+			if v.ID == variable.ID && featureDecision.Variation.FeatureEnabled {
+				featureValue = v.Value
+				break
+			}
+		}
+	}
+
+	var typeName = ""
+	var valueParsed interface{}
+	switch valueType.(type) {
```
mikeng13: Instead of passing in a variable for which we then get the type of, we can just pass in a string with the name of the type. This is less expensive to do.

optimizely/client/client.go
```diff
@@ -133,6 +135,115 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
 	return enabledFeatures, nil
 }
 
+// GetFeatureVariableString returns string feature variable value
+func (o *OptimizelyClient) GetFeatureVariableString(featureKey string, variableKey string, userContext entities.UserContext) (value string, err error) {
+	val, err := o.getFeatureVariable("", featureKey, variableKey, userContext)
+	if err == nil {
+		switch val.(type) {
```
mikeng13: Here we can
 ```
 if returnValue, ok := val.(string); ok {
  value = returnValue
}
```

Then at the end you return the named variables `value, err`

File: optimizely/client/client.go [hunk 3]
```diff
@@ -133,6 +135,115 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
...
+			return val.(string), err
+		default:
+			break
+		}
+	}
+	return "", err
+}
+
+func (o *OptimizelyClient) getFeatureVariable(valueType interface{}, featureKey string, variableKey string, userContext entities.UserContext) (value interface{}, err error) {
+	if !o.isValid {
```
mikeng13: I feel this check, along with the project config check below can likely be collapsed into a `isValid` method that takes in the method name so it can print it out if not valid

optimizely/interface.go
```diff
@@ -32,6 +33,7 @@ type ProjectConfig interface {
 	GetEventByKey(string) (entities.Event, error)
 	GetExperimentByKey(string) (entities.Experiment, error)
 	GetFeatureByKey(string) (entities.Feature, error)
+	GetFeatureFlagByKey(string) (datafileEntities.FeatureFlag, error)
```
mikeng13: This is wrong. Feature flags an d features are the same thing. You are bringing in the data file raw feature flags in here, which is not what we want.

optimizely/config/datafileprojectconfig/entities/entities.go
```diff
@@ -51,6 +55,20 @@ type FeatureFlag struct {
 	Variables     []Variable `json:"variables"`
 }
 
+// GetVariable returns variable for key from feature flag
+func (f *FeatureFlag) GetVariable(key string) (Variable, error) {
```
mikeng13: Instead of adding this functionality to the model, let's add this method to the project config itself to get a variable with key from the feature

optimizely/config/datafileprojectconfig/config.go
```diff
@@ -39,6 +40,7 @@ type DatafileProjectConfig struct {
 	experimentKeyToIDMap map[string]string
 	experimentMap        map[string]entities.Experiment
 	featureMap           map[string]entities.Feature
+	featureFlagMap       map[string]datafileEntities.FeatureFlag
```
mikeng13: See comment at bottom. This is incorrect. We should be using the entities.* which are public facing. The datafileEntities are internal models that we don't expose and are only needed so we can parse into entities.* models

optimizely/config/datafileprojectconfig/mappers/experiment.go
```diff
@@ -44,6 +44,11 @@ func mapVariation(rawVariation datafileEntities.Variation) entities.Variation {
 		Key:            rawVariation.Key,
 		FeatureEnabled: rawVariation.FeatureEnabled,
 	}
+
+	for _, variable := range rawVariation.Variables {
+		variation.Variables = append(variation.Variables, entities.VariationVariable{ID: variable.ID, Value: variable.Value})
```
mikeng13: I think we can benefit from making `Feature.Variables` a map of string to variable for easier access. It also helps that we can easily iterate over maps in Go.

optimizely/client/client_test.go
```diff
@@ -297,3 +302,216 @@ func TestGetEnabledFeaturesPanic(t *testing.T) {
...
+	testUserContext := entities.UserContext{ID: "test_user_1"}
+	testVariationVariable := entities.VariationVariable{
+		ID:    "1",
+		Value: testVariableValue,
+	}
+	testVariable := entities.Variable{
+		DefaultValue: "default",
+		ID:           "1",
+		Key:          "test_feature_flag_key",
+		Type:         "string"}
```
mikeng13: Please move the `}` to a new line

optimizely/client/client_test.go
```diff
@@ -297,3 +302,216 @@ func TestGetEnabledFeaturesPanic(t *testing.T) {
...
+	testUserContext := entities.UserContext{ID: "test_user_1"}
+	testVariationVariable := entities.VariationVariable{
+		ID:    "1",
+		Value: testVariableValue,
+	}
+	testVariable := entities.Variable{
+		DefaultValue: "default",
+		ID:           "1",
+		Key:          "test_feature_flag_key",
+		Type:         "boolean"}
```
mikeng13: Move the `}` to a new line

optimizely/client/client_test.go
```diff
@@ -297,3 +302,216 @@ func TestGetEnabledFeaturesPanic(t *testing.T) {
...
+	testUserContext := entities.UserContext{ID: "test_user_1"}
+	testVariationVariable := entities.VariationVariable{
+		ID:    "1",
+		Value: testVariableValue,
+	}
+	testVariable := entities.Variable{
+		DefaultValue: "defaultString",
+		ID:           "1",
+		Key:          "test_feature_flag_key",
+		Type:         "string"}
```
mikeng13: ditto above

optimizely/config/datafileprojectconfig/entities/entities.go
```diff
@@ -67,10 +67,16 @@ type trafficAllocation struct {
 
 // Variation represents an experiment variation from the Optimizely datafile
 type Variation struct {
-	ID string `json:"id"`
-	// @TODO(mng): include variables
-	Key            string `json:"key"`
-	FeatureEnabled bool   `json:"featureEnabled"`
+	ID             string              `json:"id"`
+	Variables      []VariationVariable `json:"variables"`
```
mikeng13: I think we can use a map here to make look-ups easier.

optimizely/event/processor_test.go
```diff
@@ -16,7 +16,7 @@ func TestDefaultEventProcessor_ProcessImpression(t *testing.T) {
 
 	assert.Equal(t, 1, processor.EventsCount())
 
-	time.Sleep(200 * time.Millisecond)
+	time.Sleep(2000 * time.Millisecond)
```
mikeng13: Is this because the test is flaky?

yasirfolio3: yes, i had instances where it failed, increasing the timeout made it more stable.

optimizely/config/datafileprojectconfig/entities/entities.go
yasirfolio3: I have updated the public facing dataModel with a map.
https://github.com/optimizely/go-sdk/pull/71/files#diff-f55225b09df17f6c9183261f39659ee1R22

optimizely/client/client.go
```diff
@@ -133,6 +122,106 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
...
+// GetFeatureVariableString returns string feature variable value
+func (o *OptimizelyClient) GetFeatureVariableString(featureKey string, variableKey string, userContext entities.UserContext) (value string, err error) {
+	val, err := o.getFeatureVariable("string", featureKey, variableKey, userContext)
+	if returnValue, ok := val.(string); ok {
+		value = returnValue
+	}
+	return value, err
+}
+
+func (o *OptimizelyClient) getFeatureVariable(valueType string, featureKey string, variableKey string, userContext entities.UserContext) (value interface{}, err error) {
```
pawels-optimizely: I think conceptually it will be better if that function returns the string instead of interface and GetFeatureVariableString does the checking. That way we eliminate valueType parameter.

pawels-optimizely: actually GetFeatureVariableString will not have to do any checking - just pass-through, but GetFeatureVariableBool, GetFeatureVariableFloat, etc  will do some checking

mikeng13: That is a great point. I think we should do it that way even if it ends up duplicating some of the code at the bottom since strong typing does give us other kinds of guarantees @yasirfolio3

yasirfolio3: @pawels-optimizely we need valueType parameter for this condition
https://github.com/optimizely/go-sdk/pull/71/files#diff-b852811f33e54bd7b6e89917e41eefadR203

yasirfolio3: @mikeng13 yes we move the checking part to each individual api such as GetFeatureVariableString, but we still do need valueType argument to check if variable.Type is the required valueType, any thoughts?

pawels-optimizely: my understanding is that you will have GetFeatureVariableString, GetFeatureVariableBool, GetFeatureVariableFloat - then they can check (return error) if a string cannot be converted to a specific type. Unless I am missing something...

yasirfolio3: > my understanding is that you will have GetFeatureVariableString, GetFeatureVariableBool, GetFeatureVariableFloat - then they can check (return error) if a string cannot be converted to a specific type. Unless I am missing something...

yes that's correct, what about the valueType parameter? we would still need it for this condition
https://github.com/optimizely/go-sdk/pull/71/files#diff-b852811f33e54bd7b6e89917e41eefadR203

yasirfolio3: > I think conceptually it will be better if that function returns the string instead of interface and GetFeatureVariableString does the checking. That way we eliminate valueType parameter.

pointing to your suggestion about elimination valueType parameter.

pawels-optimizely: that can be checked by individual function: GetFeatureVariableBool will check for bool, GetFeatureVariableFloat will check for float, etc. All you care about is whether you get an error or not when you try to convert to a particular type.

yasirfolio3: what about GetFeatureVariableString? since variable.Value is a string type, it will find any value as a valid string and return it. for e.g:
if variable.Value = "2" and variable.Type="integer", then GetFeatureVariableString should be returning an error. But since we are talking about removing valueType parameter, we will also have to remove the condition value.Type == valueType, which will cause GetFeatureVariableString to return any kind of value as a valid string.

pawels-optimizely: you are calling it as getFeatureVariable("string", featureKey, variableKey, userContext), that means variable "2" would be treated as a string in GetFeatureVariableString

pawels-optimizely: also if you need the type, you can return a string and the type and check that in GetFeatureVariableString

yasirfolio3: > also if you need the type, you can return a string and the type and check that in GetFeatureVariableString

il go ahead with this idea!

pawels-optimizely: yeah, sorry to be a pain - I am not 100% on this code yet, but it's kind of awkward flow when I read the code: 1. getting a string, 2. converting to interface, 3. returning the interface, 4. converting interface to string again - this is your overall flow. Just trying to avoid that.

yasirfolio3: no worries, can you review it now!

optimizely/client/client.go
```diff
@@ -133,6 +121,82 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
 	return enabledFeatures, nil
 }
 
+// GetFeatureVariableString returns string feature variable value
+func (o *OptimizelyClient) GetFeatureVariableString(featureKey string, variableKey string, userContext entities.UserContext) (value string, err error) {
```
pawels-optimizely: nice !, much cleaner now, no hardcoded parameter, more readable, no unneeded interface. small change but cleaned a lot of things and eliminated confusion. lgtm

optimizely/client/client.go
```diff
@@ -133,6 +121,82 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
...
+	}
+
+	featureDecision, err := o.decisionService.GetFeatureDecision(featureDecisionContext, userContext)
+	if err == nil {
+		if v, ok := featureDecision.Variation.Variables[variable.ID]; ok && featureDecision.Variation.FeatureEnabled {
+			featureValue = v.Value
+		}
+	}
+
+	if featureValue == "" {
```
mikeng13: Is this still needed? Empty strings should be OK. I think we should revert the above flow where if `err != nil` then we return this error message

yasirfolio3: done! i'l do that for other getFeatureVariable API's, getFeatureVariableString doesn't require type-casting so i don't think we need it.

yasirfolio3:
> @yasirfolio3 can you please resolve merge conflicts?

yep, done!

PR #72, opened 2019-08-22
(chore): add a track call

thomaszurkan-optimizely:


optimizely/client/client.go
```diff
@@ -133,6 +133,35 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
...
+		}
+	}()
+
+	configEvent, err := o.configManager.GetConfig().GetEventByKey(eventKey)
+
+	if err == nil {
+		userEvent := event.CreateConversionUserEvent(o.configManager.GetConfig(), configEvent, userContext, eventTags)
+		o.eventProcessor.ProcessEvent(userEvent)
+	} else {
+		logger.Error("Error getting event", err)
```
mikeng13: Please include the event key in the message

optimizely/client/client.go
```diff
@@ -133,6 +133,35 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
...
+			err := errors.New(errorMessage)
+			logger.Error(errorMessage, err)
+		}
+	}()
+
+	configEvent, err := o.configManager.GetConfig().GetEventByKey(eventKey)
+
+	if err == nil {
+		userEvent := event.CreateConversionUserEvent(o.configManager.GetConfig(), configEvent, userContext, eventTags)
+		o.eventProcessor.ProcessEvent(userEvent)
```
mikeng13: Nice and easy!

optimizely/client/client.go
```diff
@@ -133,6 +133,35 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
 	return enabledFeatures, nil
 }
 
+func (o *OptimizelyClient) Track(userContext entities.UserContext, eventKey string, eventTags map[string]interface{}) error {
+	if !o.isValid {
+		errorMessage := "Optimizely instance is not valid. Failing GetEnabledFeatures."
+		err := errors.New(errorMessage)
+		logger.Error(errorMessage, nil)
```
pawels-optimizely: err is shadowed , logger.Error takes nil ? 
something like this could be better:

err = errors.New("Optimizely instance is not valid. Failing GetEnabledFeatures.")
logger.Error("",err) or logger.Error(err.String(), err) // kind of redundant 
return err

optimizely/client/client.go
```diff
@@ -133,6 +133,35 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
...
+		return err
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
+			errorMessage := fmt.Sprintf(`Optimizely SDK is panicking with the error "%s"`, string(debug.Stack()))
+			err := errors.New(errorMessage)
```
pawels-optimizely: do you want to return that err ? or just created for logger ?

thomaszurkan-optimizely: it's in a defer

pawels-optimizely: it's not going to change the value of the error returned that's why I am asking

mikeng13: Ah he's right @thomaszurkan-optimizely, you need to name the variable in the return part of the function signature to `(err error)` so that you can assign `err = errors.New(errorMessage` instead of using `:=`

PR #82, opened 2019-08-29
Add external access to the OptimizelyClient ProjectConfig.

mikecdavis:
## Summary
- Update `ProjectConfigManager#GetConfig` also return err
- Update `ProjectConfigManager` implementations with the new signature
- Replace `isClientValid` with standard `GetProjectConfig`
- Slight refactor of the `MockProjectConfigManager` to be a bit more flexible
- Add tests to maintain coverage

When exposing ProjectConfig I felt that the use of `isValidClient` was better served with the "comma-err" idiom. So I replaced that initial check with a common method `GetProjectConfig` with is used internally and externally to get a project config.

### DIFF
--- a/optimizely/client/client.go
+++ b/optimizely/client/client.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"reflect"
 	"runtime/debug"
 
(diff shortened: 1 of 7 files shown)


mikecdavis:
Closing in favor of https://github.com/optimizely/go-sdk/pull/84

PR #83, opened 2019-08-29
(fix): first pass at event queued event dispatcher

thomaszurkan-optimizely:


optimizely/event/dispatcher.go
```diff
@@ -20,11 +23,26 @@ type Dispatcher interface {
...
+func (ed *QueueEventDispatcher) DispatchEvent(event LogEvent, callback func(success bool)) {
+
+	ed.q.Add(event)
+	callback(true)
+	go func() {
+		ed.flushEvents()
+	}()
+}
 
+func dispatchEvent(event LogEvent, callback func(success bool)) {
```
mikeng13: Having this static method might make it hard to unit test. Maybe we can try having the QueueEventDispatcher wrap around the HttpEventDispatcher and call the HttpEventDispatcher when it flushes?

optimizely/event/dispatcher.go
```diff
@@ -38,10 +56,48 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(success
...
+				time.Sleep(5 * time.Second)
+			}
+		})
+	}
 }
+
+func NewQueueEventDispatcher() Dispatcher {
+	dispatcher := &QueueEventDispatcher{q:NewInMemoryQueue(1000)}
+	return dispatcher
+}
```
mikeng13: nit: newline

optimizely/event/dispatcher.go
```diff
@@ -20,11 +23,26 @@ type Dispatcher interface {
...
+
+type QueueEventDispatcher struct {
+	q Queue
+	mux sync.Mutex
+}
+
+func (ed *QueueEventDispatcher) DispatchEvent(event LogEvent, callback func(success bool)) {
+
+	ed.q.Add(event)
+	callback(true)
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: for the queued dispatch event, I have it calling the callback once it is queued.  the idea being that eventually, it would be sent and it is no longer the responsibility of the event processor.

PR #84, opened 2019-08-30
Replace isValidClient with GetProjectConfig.

mikecdavis:
## Summary
- Update `ProjectConfigManager#GetConfig` also return err
- Update `ProjectConfigManager` implementations with the new signature
- Replace `isClientValid` with standard `GetProjectConfig`
- Slight refactor of the `MockProjectConfigManager` to be a bit more flexible
- Add tests to maintain coverage

When exposing ProjectConfig I felt that the use of `isValidClient` was better served with the "comma-err" idiom. So I replaced that initial check with a common method `GetProjectConfig` with is used internally and externally to get a project config.

Note: this replaces https://github.com/optimizely/go-sdk/pull/82

### DIFF
--- a/optimizely/client/client_test.go
+++ b/optimizely/client/client_test.go
@@ -54,12 +54,23 @@ func (c *MockProjectConfig) GetEventByKey(string) (entities.Event, error) {
 }
 
 type MockProjectConfigManager struct {
+	projectConfig optimizely.ProjectConfig
 	mock.Mock
 }
 
-func (p *MockProjectConfigManager) GetConfig() optimizely.ProjectConfig {
+func (p *MockProjectConfigManager) GetConfig() (optimizely.ProjectConfig, error) {
+	if p.projectConfig != nil {
+		return p.projectConfig, nil
+	}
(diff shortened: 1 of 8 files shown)


optimizely/client/client_test.go
```diff
@@ -210,7 +209,7 @@ func TestIsFeatureEnabled(t *testing.T) {
 	mockConfig := new(MockProjectConfig)
 	mockConfig.On("GetFeatureByKey", testFeatureKey).Return(testFeature, nil)
 	mockConfigManager := new(MockProjectConfigManager)
-	mockConfigManager.On("GetConfig").Return(mockConfig)
+	mockConfigManager.On("GetConfig").Return(mockConfig, (error)(nil))
```
mikeng13: Any reason why you have to cast the nil using `(error) (nil)`? I thought just returning `nil` would work

mikecdavis: I could have sworn it was giving be an error without it, but not now. Making this change, then merging.

PR #85, opened 2019-08-30
feat(factory-ep): Added EventProcessor to factory and made fixes to run FSC feature-management tests.

yasirfolio3:
# Summary
- Added `EventProcessor` to factory options.
- Add `batchSize` as a parameter in `NewEventProcessor` method
- Added condition to check experiment status in `CompositeExperimentService`.
- Updated dataModels as per FSC tests requirements.

# Tests
- Unit tests in `composite_experiment_service_test.go`


optimizely/event/processor.go
```diff
@@ -45,14 +45,18 @@ type QueueingEventProcessor struct {
...
 // NewEventProcessor returns a new instance of QueueingEventProcessor with queueSize and flushInterval
-func NewEventProcessor(ctx context.Context, queueSize int, flushInterval time.Duration) *QueueingEventProcessor {
+func NewEventProcessor(ctx context.Context, batchSize, queueSize int, flushInterval time.Duration) *QueueingEventProcessor {
 	p := &QueueingEventProcessor{
 		MaxQueueSize:    queueSize,
 		FlushInterval:   flushInterval,
 		Q:               NewInMemoryQueue(queueSize),
 		EventDispatcher: &HTTPEventDispatcher{},
 	}
 	p.BatchSize = 10
```
thomaszurkan-optimizely: default batch size.

yasirfolio3: @thomaszurkan-optimizely  using DefaultBatchSize required us to import client package in processor.go which was causing import cycle issue! To avoid that, i have now moved DefaultBatchSize,DefaultEventQueueSize and DefaultEventFlushInterval to processor.go. Will be committing it with some decision service test-cases to improve coverage.

yasirfolio3: @thomaszurkan-optimizely done!

optimizely/client/factory.go
```diff
@@ -37,6 +36,7 @@ type Options struct {
 	Context              context.Context
 	ProjectConfigManager optimizely.ProjectConfigManager
 	DecisionService      decision.Service
+	EventProcessor       *event.QueueingEventProcessor
```
mikeng13: This is wrong. It should reference the `Processor` interface and not the struct implementation

optimizely/client/factory.go
```diff
@@ -91,10 +88,10 @@ func (f OptimizelyFactory) ClientWithOptions(clientOptions Options) (*Optimizely
 	} else {
 		// if no context is provided, we create our own cancellable context and hand it over to the client so the client can shut down its child processes
 		ctx = context.Background()
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithCancel(ctx)
-		client.cancelFunc = cancel
 	}
+	var cancel context.CancelFunc
```
mikeng13: I don't think this should be moved here because if the user passes in a context, they should take care of making it cancelable. If the testapp needs to be able to cancel it, it should pass in a cancelable context and invoke the cancel func from outside the SDK. @pawels-optimizely does this make sense?

optimizely/decision/composite_experiment_service.go
```diff
@@ -44,6 +47,13 @@ func NewCompositeExperimentService() *CompositeExperimentService {
 
 // GetDecision returns a decision for the given experiment and user context
 func (s CompositeExperimentService) GetDecision(decisionContext ExperimentDecisionContext, userContext entities.UserContext) (ExperimentDecision, error) {
+
+	if decisionContext.Experiment.Status != entities.Running {
```
mikeng13: In the new world, experiment status is no longer relevant. In fact we should probably deprecate the experiment status tests (i.e. run them in 2.x only) because we've changed the datafile builds to only including running experiments.

optimizely/decision/composite_service.go
```diff
@@ -69,6 +69,9 @@ func (s CompositeService) GetFeatureDecision(featureDecisionContext FeatureDecis
 				"source":          featureDecision.Source,
 			},
 		}
+		if userContext.Attributes == nil {
```
mikeng13: I think nil shouldn't be an issue for us because we don't ever attempt to write to it from the SDK. It is a read-only value and we should treat it as such. No need to initialize to an empty map. @pawels-optimizely ^^

optimizely/event/events.go
```diff
@@ -62,8 +62,8 @@ type ConversionEvent struct {
 
 // LogEvent represents a log event
 type LogEvent struct {
-	endPoint string
-	event    Batch
+	EndPoint string
```
mikeng13: Nice, this should definitely be public if we expect the event dispatcher to be replaceable

optimizely/event/processor.go
```diff
@@ -42,17 +42,30 @@ type QueueingEventProcessor struct {
 	EventDispatcher Dispatcher
 }
 
+// DefaultBatchSize holds the default value for the batch size
+const DefaultBatchSize = 10
+
+// DefaultEventQueueSize holds the default value for the event queue size
+const DefaultEventQueueSize = 10
```
mikeng13: The default value should be 100

optimizely/decision/composite_service.go
yasirfolio3: FSC expect's attributes to be empty dict in case of null, il just move this check on the test-app side then.

optimizely/event/processor.go
yasirfolio3: @mikeng13 Recommended changes made.

optimizely/config/datafileprojectconfig/mappers/experiment.go
```diff
@@ -65,9 +65,10 @@ func mapExperiment(rawExperiment datafileEntities.Experiment) entities.Experimen
 	experiment := entities.Experiment{
 		AudienceIds:           rawExperiment.AudienceIds,
 		ID:                    rawExperiment.ID,
+		LayerID:               rawExperiment.LayerID,
```
mikeng13: This doesn't seem to be used. Can you remove it?

optimizely/decision/composite_service.go
```diff
@@ -62,13 +62,19 @@ func (s CompositeService) GetFeatureDecision(featureDecisionContext FeatureDecis
 
 	// @TODO: add errors
 	if s.notificationCenter != nil {
+		featureMap := map[string]interface{}{
```
mikeng13: nit: Call this `featureInfo`

optimizely/config/datafileprojectconfig/mappers/experiment.go
yasirfolio3: done!

PR #88, opened 2019-09-04
(feat): Queue event dispatcher

thomaszurkan-optimizely:
create a event dispatcher that holds onto the batch event and will try and send/resend.  This allows the processor to create a batch event and send it to the event dispatcher and remove the events from its queue.  Relying on the event dispatcher to send.


optimizely/event/dispatcher.go
```diff
@@ -54,10 +55,70 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(bool)) {
...
+			dispatcherLogger.Error(fmt.Sprintf("http.Post invalid response %d", resp.StatusCode), nil)
 			success = false
 		}
 	}
 	callback(success)
+}
+
+type QueueEventDispatcher struct {
+	eventQueue Queue
+	mux        sync.Mutex
```
mikeng13: nit: Call this eventQueueLock for clarity

optimizely/event/dispatcher.go
```diff
@@ -54,10 +55,70 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(bool)) {
...
+
+	for ed.eventQueue.Size() > 0 {
+		items := ed.eventQueue.Get(1)
+		if len(items) == 0 {
+			time.Sleep(5 * time.Second)
+			continue
+		}
+		event, ok := items[0].(LogEvent)
+		if !ok {
+			//remove it?
```
mikeng13: Is this still a question?

optimizely/event/dispatcher.go
```diff
@@ -54,10 +55,70 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(bool)) {
...
+			dispatcherLogger.Error("invalid type passed to event dispatcher", nil)
+			ed.eventQueue.Remove(1)
+			continue
+		}
+
+		ed.dispatcher.DispatchEvent(event, func(success bool) {
+			if success {
+				ed.eventQueue.Remove(1)
+			} else {
+				// we failed.  Sleep 5 seconds and try again.  Or, should we exit and try the next time
```
mikeng13: So if another log event is added, do they both get flushed? It looks like we only flush 1 at a time

optimizely/event/processor.go
```diff
@@ -97,6 +97,10 @@ func (p *QueueingEventProcessor) StartTicker(ctx context.Context) {
 			case <-ctx.Done():
 				pLogger.Debug("Event processor stopped, flushing events.")
 				p.FlushEvents()
+				d, ok := p.EventDispatcher.(*QueueEventDispatcher)
```
mikeng13: Hmm so on ticker we also try to flush the dispatcher? I think this coupling might be more than necessary. @pawels-optimizely @mikecdavis what do you all think? What about having the queue event dispatcher have its own ticker?

thomaszurkan-optimizely: no, on done from ctx we try and flush the dispatcher.  There is no timer or a need for one

optimizely/event/dispatcher.go
thomaszurkan-optimizely: no, they are sent one at a time but all are flushed.  as in for queue.size > 0

optimizely/event/processor.go
mikeng13: Oh yeah I misread. Makes sense.

optimizely/event/dispatcher.go
```diff
@@ -54,10 +55,72 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(bool)) {
...
+		})
+	}
+}
+
+// NewQueueEventDispatcher creates a dispatcher that queues in memory and then sends via go routine.
+func NewQueueEventDispatcher(ctx context.Context) Dispatcher {
+	dispatcher := &QueueEventDispatcher{eventQueue: NewInMemoryQueue(1000), dispatcher:&HTTPEventDispatcher{}}
+
+	go func() {
+		<-ctx.Done()
```
mikeng13: Is this needed if you call flush events from the parent?

thomaszurkan-optimizely: I'm not sure it is needed.  But, if there are no events to send it is pretty much a noop.

optimizely/event/dispatcher.go
```diff
@@ -54,10 +55,72 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(bool)) {
...
+
+	ed.eventFlushLock.Lock()
+
+	defer func(){
+		ed.eventFlushLock.Unlock()
+	}()
+
+	for ed.eventQueue.Size() > 0 {
+		items := ed.eventQueue.Get(1)
+		if len(items) == 0 {
```
pawels-optimizely: why do we need to wait 5 seconds here ? why len(items) == 0 when ed.eventQueue.Size() > 0 ? are they not related ?

thomaszurkan-optimizely: i am just checking the array size after the call for safety.

thomaszurkan-optimizely: just a sanity check.  not sure we need the sleep in there.

pawels-optimizely: ok, if this is a sanity check, what should we do if len(items) == 0 ? that means something went wrong, should we report it ? stop the process? I think repeater pattern would be more elegant here, unless you want to try forever by design. I don't think sleep is needed here

optimizely/event/dispatcher.go
```diff
@@ -54,10 +55,72 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(bool)) {
...
+			ed.eventQueue.Remove(1)
+			continue
+		}
+
+		ed.dispatcher.DispatchEvent(event, func(success bool) {
+			if success {
+				ed.eventQueue.Remove(1)
+			} else {
+				// we failed.  Sleep 5 seconds and try again.  Or, should we exit and try the next time a log event is added?
+				time.Sleep(5 * time.Second)
```
pawels-optimizely: should we log this ? i think a warning  would be useful here

optimizely/event/dispatcher.go
pawels-optimizely: is canceled context going to stop event dispatcher too ?

pawels-optimizely: please look at https://github.com/optimizely/go-sdk/blob/master/optimizely/config/polling_manager.go#L54 - context will stop the whole mechanism

optimizely/event/dispatcher.go
thomaszurkan-optimizely: I took out the sleep and just continue with the expectation that size will return 0 or more and I can try again

optimizely/event/dispatcher.go
mikeng13: I think this will need to be revised to account for the wait group implementation that @pawels-optimizely is working

optimizely/event/processor.go
```diff
@@ -110,6 +110,10 @@ func (p *QueueingEventProcessor) StartTicker(ctx context.Context) {
...
+				if ok {
+					d.flushEvents()
```
mikeng13: I think `FlushEvents` should be part of the dispatcher interface. If the user provides their own, wouldn't we want to be able to flush their events on cancel?

optimizely/event/dispatcher_test.go
```diff
@@ -0,0 +1,39 @@
...
+	// its been queued
+	assert.Equal(t,1, qd.eventQueue.Size())
+
+	// give the queue a chance to run
+	time.Sleep(1 * time.Second)
+
+	// check the queue
+	assert.Equal(t,0, qd.eventQueue.Size())
+
+}
```
mikeng13: nit: Add newline

optimizely/event/dispatcher_test.go
```diff
@@ -0,0 +1,39 @@
+package event
```
mikeng13: Add license header

optimizely/event/dispatcher.go
mikeng13: It's fine to merge this in and have Pawel add the wait group here afterwards

optimizely/event/dispatcher.go
```diff
@@ -54,10 +58,83 @@ func (*HTTPEventDispatcher) DispatchEvent(event LogEvent, callback func(bool)) {
...
+		if !ok {
+			// remove it
+			dispatcherLogger.Error("invalid type passed to event dispatcher", nil)
+			ed.eventQueue.Remove(1)
+			continue
+		}
+
+		ed.dispatcher.DispatchEvent(event, func(success bool) {
+			if success {
+				dispatcherLogger.Debug("event sent")
```
mikeng13: As mentioned offline, please add more context to the log message like this: 
`dispatcherLogger.Debug(fmt.Sprintf("Dispatched log event %+v", event))`

optimizely/event/dispatcher.go
mikeng13: Use a constant for this magic number

PR #90, opened 2019-09-04
(feat): add channel queue

thomaszurkan-optimizely:
can be used for event processor and/or event dispatcher.


optimizely/event/chan_queue_test.go
```diff
@@ -0,0 +1,36 @@
+package event
```
mikeng13: Please include the license header

PR #91, opened 2019-09-04
adding go.mod support (removing vendor) and replacing encoding/json with json-iterator package.

pawels-optimizely:


.travis.yml
```diff
@@ -1,3 +1,6 @@
+env:
+  - GO111MODULE=on
+
 language: go
 go:
   - 1.12.x
```
mikeng13: I think we can consider adding 1.13.x here

go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/optimizely/go-sdk
+
+go 1.12
```
mikeng13: Should this be 13?

pawels-optimizely: I think we should stay with 1.12 for now, 1.13 has been just released. We should wait and see if there are no major problems reported by go community.

.travis.yml
pawels-optimizely: I think we should stay with 1.12 for now, 1.13 has been just released. We should wait and see if there are no major problems reported by go community.

PR #97, opened 2019-09-11
chore(readme): Add instructions for version pinning.

mikeng13:
# Summary
- Adds a note on us doing trunk-based dev and one way to pin versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,23 @@ go install
 ```
 
 NOTE:
+We practice trunk-based development, and as such our default branch, `master` might not always be the most stable. We do tag releases on Github and you can pin your installation to those particular release versions. One way to do this is to use *Go Modules* for managing external dependencies:
+
+```
+module mymodule
+
+go 1.12
+
+require (
+	github.com/optimizely/go-sdk v0.1.0
+)
+```
+
+NOTE:
 ```$sh
 go get github.com/optimizely/go-sdk/...
 ```
-or 
+or
 ```$sh
 go get github.com/optimizely/go-sdk/optimizely
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,11 +86,24 @@ cd $GOPATH/src/github.com/optimizely/go-sdk
...
+NOTE:
+We practice trunk-based development, and as such our default branch, `master` might not always be the most stable. We do tag releases on Github and you can pin your installation to those particular release versions. One way to do this is to use *Go Modules* for managing external dependencies:
+
+```
+module mymodule
+
+go 1.12
+
+require (
+	github.com/optimizely/go-sdk v0.1.0
```
mikecdavis: If the are already using go.mod in their application then they can use:
```
go mod edit -require github.com/optimizely/go-sdk@0.1.0
```

README.md
```diff
@@ -86,11 +86,24 @@ cd $GOPATH/src/github.com/optimizely/go-sdk
 go install
 ```
 
+NOTE:
+We practice trunk-based development, and as such our default branch, `master` might not always be the most stable. We do tag releases on Github and you can pin your installation to those particular release versions. One way to do this is to use *Go Modules* for managing external dependencies:
```
mikecdavis: Should we link to [blog.golang.org](https://blog.golang.org/using-go-modules) or the [github wiki](https://github.com/golang/go/wiki/Modules)?

mikeng13:
Thanks for suggestions @mikecdavis . I've incorporated them.

README.md
pawels-optimizely: Yeah, from what I see this is the best way, but it did not work for me as a package installation. I think we should discuss - as I don't have that much experience with go.mod

PR #99, opened 2019-09-11
Adjusting config manager to support a Sync method

aliabbasrizvi:
Introducing this method will allow some advanced use cases like requesting datafile be updated on demand.

### DIFF
--- a/optimizely/config/polling_manager.go
+++ b/optimizely/config/polling_manager.go
@@ -53,42 +53,40 @@ type PollingProjectConfigManager struct {
 	exeCtx utils.ExecutionCtx // context used for execution control
 }
 
-func (cm *PollingProjectConfigManager) activate(initialPayload []byte, init bool) {
+func (cm *PollingProjectConfigManager) Sync(datafile []byte) {
+	var e error
+	var code int
+	if len(datafile) == 0 {
+		datafile, code, e = cm.requester.Get()
 
-	update := func() {
-		var e error
-		var code int
-		var payload []byte
-		if init && len(initialPayload) > 0 {
-			payload = initialPayload
-		} else {
-			payload, code, e = cm.requester.Get()
-
-			if e != nil {
-				cmLogger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
-			}
+		if e != nil {
+			cmLogger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
 		}
-
-		projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(payload)
-		if err != nil {
-			cmLogger.Error("failed to create project config", err)
-		}
-
-		cm.configLock.Lock()
-		cm.projectConfig = projectConfig
-		cm.err = err
-		cm.configLock.Unlock()
 	}
+
+	projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(datafile)
+	if err != nil {
+		cmLogger.Error("failed to create project config", err)
+	}
+
+	cm.configLock.Lock()
+	cm.projectConfig = projectConfig
+	cm.err = err
+	cm.configLock.Unlock()
+}
+
+func (cm *PollingProjectConfigManager) start(initialDatafile []byte, init bool) {
 
 	if init {
-		update()
+		cm.Sync(initialDatafile)
 		return
 	}
+
 	t := time.NewTicker(cm.pollingInterval)
 	for {
 		select {
 		case <-t.C:
-			update()
+			cm.Sync([]byte{})
 		case <-cm.exeCtx.GetContext().Done():
 			cmLogger.Debug("Polling Config Manager Stopped")
 			return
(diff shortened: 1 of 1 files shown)


optimizely/config/polling_manager.go
```diff
@@ -53,42 +53,40 @@ type PollingProjectConfigManager struct {
...
-
-			if e != nil {
-				cmLogger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
-			}
-		}
-
-		projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(payload)
-		if err != nil {
-			cmLogger.Error("failed to create project config", err)
+func (cm *PollingProjectConfigManager) Sync(datafile []byte) {
```
aliabbasrizvi: We can perhaps call this SyncConfig?

optimizely/config/polling_manager.go
```diff
@@ -53,42 +53,40 @@ type PollingProjectConfigManager struct {
...
+		cmLogger.Error("failed to create project config", err)
 	}
 
+	cm.configLock.Lock()
+	cm.projectConfig = projectConfig
+	cm.err = err
+	cm.configLock.Unlock()
+}
+
+func (cm *PollingProjectConfigManager) start(initialDatafile []byte, init bool) {
```
aliabbasrizvi: Wasn't sure `activate` was the right method name.

optimizely/config/polling_manager.go
```diff
@@ -114,10 +112,10 @@ func NewPollingProjectConfigManagerWithOptions(exeCtx utils.ExecutionCtx, sdkKey
 
 	pollingProjectConfigManager := PollingProjectConfigManager{requester: requester, pollingInterval: pollingInterval, exeCtx: exeCtx}
 
-	pollingProjectConfigManager.activate(options.Datafile, true) // initial poll
+	pollingProjectConfigManager.start(options.Datafile, true) // initial poll
 
 	cmLogger.Debug("Polling Config Manager Initiated")
-	go pollingProjectConfigManager.activate([]byte{}, false)
+	go pollingProjectConfigManager.start([]byte{}, false)
```
aliabbasrizvi: This looks awkward that we call start twice. I did not change this right away, but we should change this behavior.

optimizely/config/polling_manager.go
pawels-optimizely: does not matter, activate activates polling manager, I have no problem with that name. but you can use start too, it does not matter

optimizely/config/polling_manager.go
```diff
@@ -53,42 +53,40 @@ type PollingProjectConfigManager struct {
...
+		cm.Sync(initialDatafile)
 		return
 	}
+
 	t := time.NewTicker(cm.pollingInterval)
 	for {
 		select {
 		case <-t.C:
-			update()
+			cm.Sync([]byte{})
```
pawels-optimizely: keep private, sync instead of Sync

optimizely/config/polling_manager.go
pawels-optimizely: please add a comment, otherwise linters will complain 
comment has to be like // SyncConfig .......

mikeng13: Nah, sync is fine

optimizely/config/polling_manager.go
```diff
@@ -53,42 +53,42 @@ type PollingProjectConfigManager struct {
...
-		}
-
-		projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(payload)
-		if err != nil {
-			cmLogger.Error("failed to create project config", err)
+// SyncConfig gets current datafile and updates projectConfig
+func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
+	var e error
+	var code int
+	if len(datafile) == 0 {
```
mikeng13: Can we do a nil check and pass in `nil` from the ticker sync call below? It is more accurate to treat it as nil than an empty byte array

optimizely/config/polling_manager.go
pawels-optimizely: polling_manager implies to sync datafile only, so Sync would be fine, but I will not enforce it too much

optimizely/config/polling_manager.go
```diff
@@ -53,42 +53,42 @@ type PollingProjectConfigManager struct {
...
+		if e != nil {
+			cmLogger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
 		}
+	}
 
-		cm.configLock.Lock()
-		cm.projectConfig = projectConfig
-		cm.err = err
-		cm.configLock.Unlock()
+	// TODO: Compare revision numbers here and set projectConfig only if the revision number has changed
```
mikeng13: nit: This TODO should go on line 73

optimizely/config/polling_manager.go
```diff
@@ -53,42 +53,42 @@ type PollingProjectConfigManager struct {
...
+		cm.SyncConfig(initialDatafile)
 		return
 	}
+
 	t := time.NewTicker(cm.pollingInterval)
 	for {
 		select {
 		case <-t.C:
-			update()
+			cm.SyncConfig([]byte{})
```
mikeng13: Pass in nil instead

pawels-optimizely: For now it can be fine (passing bytes or nil), but I think CM should eventually have this info.
In my opinion, both sdk key and datafile url have to be the properties of CM. so SyncConfig will need to change to something like PollAndSync, and we can have a similar method (StaticSync) with provided bytes.

optimizely/config/polling_manager.go
pawels-optimizely: we will be able to pass both, so we need to check for len of datafile when it's not nil, and also for nil itself

aliabbasrizvi: I am going by @pawels-optimizely 's comment and leaving this as is.

optimizely/config/polling_manager.go
aliabbasrizvi: I am going by @pawels-optimizely 's comment and leaving this as is.

PR #102, opened 2019-09-12
added requester to the dispatcher, and providing with highly customizable client

pawels-optimizely:
1. move requester to util, added post requests, and using that in dispatcher (now we have timeout on the client)
2. Implemented functional optional parameter on client, added example with custom polling interval and some other parameters for event processor.


optimizely/client/factory.go
```diff
@@ -43,6 +42,47 @@ type OptimizelyFactory struct {
...
+		executionCtx: executionCtx,
+	}
+	for _, opt := range clientOptions {
+		opt(appClient, executionCtx)
+	}
+	return appClient
+}
+
+// ConfigManager sets config manager on a client
+func ConfigManager(sdkKey string, pollingInterval time.Duration, dataFile []byte) OptionFunc {
```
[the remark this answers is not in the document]
mikeng13: Ok I can see the.merit in keeping it simple without the struct options. Here are the chances needed:

- ConfigManager func should take in a config manager interface and set it on the instance
- PollingConfigManager func should take in the current args for ConfigManager
- EventProcessor func should take in an event processor interface type
- BatchEventProcessor func should take in the batch options you have for the current EventProcessor func
- DecisionService func should take in a decision service interface type
- CompositeDecisionService can take in a notification center

optimizely/client/factory.go
```diff
@@ -43,6 +42,47 @@ type OptimizelyFactory struct {
...
 }
 
+// OptionFunc is a type to a proper func
+type OptionFunc func(*OptimizelyClient, utils.ExecutionCtx)
+
+// GetClient gets client and sets some parameters
+func (f OptimizelyFactory) GetClient(clientOptions ...OptionFunc) *OptimizelyClient {
+	executionCtx := utils.NewCancelableExecutionCtx()
+	appClient := &OptimizelyClient{
+		isValid:      false,
```
mikeng13: This should be true right? Otherwise the client is unusable

optimizely/client/factory.go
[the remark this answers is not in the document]
pawels-optimizely: done

optimizely/client/factory.go
```diff
@@ -43,6 +42,68 @@ type OptimizelyFactory struct {
...
+		executionCtx: executionCtx,
+	}
+	for _, opt := range clientOptions {
+		opt(appClient, executionCtx)
+	}
+	return appClient
```
mikeng13: If we arrive at this step we should set isvalid to true for the client to be usable.

optimizely/client/factory.go
pawels-optimizely: I see, the question is what are the cases that make the client invalid ? What is the minimum requirement to make the client valid? I don't want to set it blindly to true, otherwise we don't need that variable at all.

optimizely/client/factory.go
mikeng13: As we discussed offline, let's get rid of the `isValid` property as it does not make sense to keep right now.

pawels-optimizely: done please check

PR #103, opened 2019-09-12
Add unit tests for datafile project config.

mikecdavis:
This adds unit tests for all of the config getters.


optimizely/config/datafileprojectconfig/config_test.go
```diff
@@ -0,0 +1,333 @@
...
+		featureMap: featureMap,
+	}
+
+	actual, err := config.GetFeatureByKey(key)
+	assert.Nil(t, err)
+	assert.Equal(t, feature, actual)
+}
+
+func TestGetFeatureByKeyWithError(t *testing.T) {
+	config := &DatafileProjectConfig{}
```
mikeng13: Can we assert the err message also?

PR #104, opened 2019-09-13
Added info log for datafile revision update.

yasirfolio3:
1. Added datafile revision updated log in datafile polling manager.

### DIFF
--- a/optimizely/config/polling_manager_test.go
+++ b/optimizely/config/polling_manager_test.go
@@ -105,7 +105,7 @@ func TestNewPollingProjectConfigManagerWithSimilarDatafileRevisions(t *testing.T
 
 func TestNewPollingProjectConfigManagerWithDifferentDatafileRevisions(t *testing.T) {
 	mockDatafile1 := []byte(`{"revision":"42","projectId":"1000"}`)
-	mockDatafile2 := []byte(`{"revision":"43","projectId":"1001"}`)
+	mockDatafile2 := []byte(`{"revision":"43","projectId":"1000"}`)
 	projectConfig1, _ := datafileprojectconfig.NewDatafileProjectConfig(mockDatafile1)
 	projectConfig2, _ := datafileprojectconfig.NewDatafileProjectConfig(mockDatafile2)
 	mockRequester := new(MockRequester)
--- a/optimizely/config/polling_manager_test.go
+++ b/optimizely/config/polling_manager_test.go
@@ -78,8 +78,8 @@ func TestNewPollingProjectConfigManagerWithNull(t *testing.T) {
 }
 
 func TestNewPollingProjectConfigManagerWithSimilarDatafileRevisions(t *testing.T) {
-	mockDatafile1 := []byte(`{"revision":"42","projectId":"1000"}`)
-	mockDatafile2 := []byte(`{"revision":"42","projectId":"1001"}`)
+	mockDatafile1 := []byte(`{"revision":"42","botFiltering":true}`)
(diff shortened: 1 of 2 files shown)


yasirfolio3:
@mikeng13 i have spotted an issue, The test's for polling manager defined here `https://github.com/optimizely/go-sdk/blob/4457b50f3d7da827143667b5dfafd07c1cc6d7bc/optimizely/config/polling_manager_test.go#L39` are not working correctly as NewDatafileProjectConfig method is returning nil config for the mock datafile. They are passing because `configManager.GetConfig` is also returning nil for this test.

yasirfolio3:
> Can we add unit tests for config manager and check status of project config through these various scenarios?

Done, i have also fixed one of the previous tests which was producing nil project config due to unmarshalling issues, can you please review it now?

optimizely/config/polling_manager_test.go
```diff
@@ -74,3 +76,57 @@ func TestNewPollingProjectConfigManagerWithNull(t *testing.T) {
 	_, err := configManager.GetConfig()
 	assert.NotNil(t, err)
 }
+
+func TestNewPollingProjectConfigManagerWithSimilarDatafileRevisions(t *testing.T) {
+	mockDatafile1 := []byte(`{"revision":"42","projectId":"1000"}`)
+	mockDatafile2 := []byte(`{"revision":"42","projectId":"1001"}`)
```
aliabbasrizvi: Ah, this is interesting. I'd recommend using the same `projectId` or better create 2 project configs from the same mock datafile. So you can just have 1 variable `mockDatafile` and create `projectConfig1` and `projectConfig2` from it.

optimizely/config/polling_manager_test.go
```diff
@@ -74,3 +76,57 @@ func TestNewPollingProjectConfigManagerWithNull(t *testing.T) {
...
+	assert.NotNil(t, actual)
+	assert.Equal(t, projectConfig1, actual)
+
+	configManager.SyncConfig(mockDatafile2)
+	actual, err = configManager.GetConfig()
+	assert.Equal(t, projectConfig1, actual)
+}
+
+func TestNewPollingProjectConfigManagerWithDifferentDatafileRevisions(t *testing.T) {
+	mockDatafile1 := []byte(`{"revision":"42","projectId":"1000"}`)
```
aliabbasrizvi: Same as above.

optimizely/config/polling_manager_test.go
yasirfolio3: A slight change in both mock datafile's was required to verify that after calling `SyncConfig` using the second mock datafile, we still get project config generated for the first mock datafile. As per the suggestion, if we created 2 configs using the same mockDatafile, we won't be able to distinguish whether config for first datafile was returned or the second?

optimizely/config/polling_manager_test.go
yasirfolio3: I have updated the projectId for this test.

optimizely/config/polling_manager_test.go
aliabbasrizvi: My problem is with the premise of the test itself. It is an improbable event for project ID to be different. It does not line up with the mental model.

Why not pick a different field. Say set `botFiltering` to `False` in one datafile and to `True` in other datafile.

yasirfolio3: Thanks, makes complete sense, i have made the suggested changes in the test-case.

PR #105, opened 2019-09-13
feat(notification-listener): Added support for removing notification listener.

yasirfolio3:


yasirfolio3:
> We actually need to return the notification id from OnDecision and use pass that id to RemoveOnDecision so that we don't end up removing all the handlers

done, can you review again?

optimizely/notification/manager.go
```diff
@@ -24,6 +24,7 @@ import (
 // Manager is a generic interface for managing notifications of a particular type
 type Manager interface {
 	AddHandler(func(interface{})) (int, error)
+	RemoveHandler(id int)
```
mikeng13: Please add unit tests for this method in the manager

yasirfolio3: Thanks for pointing out, done!

optimizely/notification/center.go
```diff
@@ -49,6 +50,16 @@ func (c *DefaultCenter) AddHandler(notificationType Type, handler func(interface
 	return -1, fmt.Errorf("no notification manager found for type %s", notificationType)
 }
 
+// RemoveHandler removes a handler for the given id and notification type
+func (c *DefaultCenter) RemoveHandler(id int, notificationType Type) error {
```
pawels-optimizely: if you have a better name for RemoveHandler, then please use it. we have that name on manager already. (I am not really enforcing - just suggesting). otherwise lgtm

yasirfolio3: thanks for the feedback, any suggestions?

yasirfolio3: We also have AddHandler method on both notificationCenter and manager, we will have to change that aswell then?

pawels-optimizely: to remove confusion, I would probably remove "Handler" word in notification package. I would have just Add and Remove. Since this is notification package, we know we are adding and removing a notification.  But like I said, I will not enforce it. I have just seen a lot of cases of overusing methods with same name for different objects.

yasirfolio3: i think so too that `Add` and `Remove` make more sense, have made these changes, can you please review it now?

pawels-optimizely: are we adding a notification, or a decision ? I think you did the other way around

yasirfolio3: we are adding notification listener for any notification type, but currently we have only `Decision` type as of now in go. I have updated the method definition for notification-manager now instead of notification-center. Can you please re-review it?

optimizely/notification/manager.go
```diff
@@ -40,13 +41,21 @@ func NewAtomicManager() *AtomicManager {
...
+// Add adds the given handler
+func (am *AtomicManager) Add(newHandler func(interface{})) (int, error) {
 	atomic.AddUint32(&am.counter, 1)
 	am.handlers[am.counter] = newHandler
 	return int(am.counter), nil
 }
 
+// Remove removes handler with the given id
+func (am *AtomicManager) Remove(id int) {
+	counter := uint32(id)
```
mikeng13: nit: This isn't really a counter. Let's call it `handlerID`

yasirfolio3: fixed.

PR #106, opened 2019-09-16
Moving experiment evaluation from CompositeFeatureService to FeatureExperimentService

aliabbasrizvi:
This is in preparation for adding mutex support.

### DIFF
--- a/optimizely/decision/composite_feature_service.go
+++ b/optimizely/decision/composite_feature_service.go
@@ -28,15 +28,15 @@ var cfLogger = logging.GetLogger("CompositeFeatureService")
 
 // CompositeFeatureService is the default out-of-the-box feature decision service
 type CompositeFeatureService struct {
-	featureExperimentDecisionService ExperimentService
-	rolloutDecisionService    FeatureService
+	featureExperimentService ExperimentService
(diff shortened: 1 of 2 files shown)


optimizely/decision/feature_experiment_service.go
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package decision //
+package decision
+
+import (
+	"github.com/optimizely/go-sdk/optimizely/entities"
+)
+
```
pawels-optimizely: it needs a comment - linters failing

optimizely/decision/composite_feature_service.go
```diff
@@ -28,14 +28,14 @@ var cfLogger = logging.GetLogger("CompositeFeatureService")
 
 // CompositeFeatureService is the default out-of-the-box feature decision service
 type CompositeFeatureService struct {
-	experimentDecisionService ExperimentService
+	featureExperimentDecisionService ExperimentService
```
mikeng13: nit: We don't need the word `Decision` as it is redundant. So `featureExperimentService`

PR #107, opened 2019-09-17
Introducing mutex support for feature tests

aliabbasrizvi:
This change introduces support for feature tests belonging to the same mutex group.

NOTE: The current change avoids refactoring a lot of the code as the ideal way to do this is to bucket into group to determine experiment and then bucket into experiment. 
For eg. let's say there are N experiments in a mutex group. In the current scenario we will bucket N times to evaluate feature tests, but we only need to bucket 2 times: once for group and once for experiment.

That will come as an enhancement to this code.

### DIFF
--- a/optimizely/decision/bucketer/experiment_bucketer.go
+++ b/optimizely/decision/bucketer/experiment_bucketer.go
@@ -56,8 +56,8 @@ func (b MurmurhashBucketer) Bucket(bucketingID string, experiment entities.Exper
 		bucketKey := bucketingID + group.ID
 		bucketedExperimentID := b.bucketToEntity(bucketKey, group.TrafficAllocation)
 		if bucketedExperimentID == "" || bucketedExperimentID != experiment.ID {
-			// User is not bucketed into an experiment in the exclusion group, return nil variation
-			return nil, reasons.NotInGroup, nil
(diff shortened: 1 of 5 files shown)


optimizely/decision/bucketer/experiment_bucketer.go
```diff
@@ -56,8 +56,8 @@ func (b MurmurhashBucketer) Bucket(bucketingID string, experiment entities.Exper
 		bucketKey := bucketingID + group.ID
 		bucketedExperimentID := b.bucketToEntity(bucketKey, group.TrafficAllocation)
 		if bucketedExperimentID == "" || bucketedExperimentID != experiment.ID {
-			// User is not bucketed into an experiment in the exclusion group, return nil variation
-			return nil, reasons.NotInGroup, nil
+			// User is not bucketed into provided experiment in mutex group
+			return nil, reasons.NotBucketedIntoVariation, nil
```
mikeng13: Why are we changing it from the `NotInGroup` message?

aliabbasrizvi: Not in group is incorrect because you may be in another experiment in the group. You are not in this particular experiment and hence `NotBucketedIntoVariation` is correct and not `NotInGroup`

PR #110, opened 2019-09-19
handling recovery errors more elegantly

pawels-optimizely:
one comment:  I tried to put that recovery logic in a common function and that did not work. For some reason recovery did not initiate from outside the calling function.

### DIFF
--- a/optimizely/client/client.go
+++ b/optimizely/client/client.go
@@ -76,9 +76,18 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext) 
 
 	defer func() {
 		if r := recover(); r != nil {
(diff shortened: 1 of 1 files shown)


optimizely/client/client.go
```diff
@@ -76,9 +76,17 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
 
 	defer func() {
 		if r := recover(); r != nil {
-			errorMessage := fmt.Sprintf(`optimizely SDK is panicking with the error "%s"`, string(debug.Stack()))
-			err = errors.New(errorMessage)
+			switch t := r.(type) {
```
mikeng13: We need this in `IsFeatureEnabled` as well as the `GetFeatureVariable*` methods.

PR #112, opened 2019-09-20
client is not blocking now, waits for polling config to set up the file.

pawels-optimizely:


optimizely/config/polling_manager.go
```diff
@@ -139,6 +136,26 @@ func NewPollingProjectConfigManager(exeCtx utils.ExecutionCtx, sdkKey string) *P
...
+	startTs := time.Now()
+	for time.Now().Before(startTs.Add(defaultWaitConfig)) {
+		cm.configLock.RLock()
+		if cm.projectConfig != nil {
+			defer cm.configLock.RUnlock()
+			return cm.projectConfig, cm.err
+		}
+		cm.configLock.RUnlock()
+		cmLogger.Debug("Waiting for Polling Manager to be initialized")
+		time.Sleep(time.Second)
```
mjc1283: I'm not very qualified to review this, but two things seem less than ideal:
- Lots of locking & unlocking
- Must wait until the next second boundary after config is fetched

Why can't we use 2 channels, and on one we write a value after 10 seconds signaling timeout, and on the other we write the config after it's fetched, then select on those 2 channels?

mikeng13:
After syncing offline, we've decided not to continue pursuing an async initialization method for now as the amount of effort and complexity might not be worth the effort. @pawels-optimizely I am gonna close this PR.

PR #113, opened 2019-09-23
(feat): add a way to create a custom event processor taking advantage of Exec…

thomaszurkan-optimizely:
…utionContext

This PR alters the NewEventProcessor so that you pass in EPOptionConfig for batchsize, flush interval, queue size, as well as queue or dispatcher.  This will be used in side door to create a NSQ based queue used by the event processor.  It will also allow for side door event processor that forwards to side door to be used as well.

### DIFF
--- a/optimizely/event/processor.go
+++ b/optimizely/event/processor.go
@@ -58,11 +58,23 @@ var pLogger = logging.GetLogger("EventProcessor")
 
 // NewEventProcessor returns a new instance of QueueingEventProcessor with queueSize and flushInterval
 func NewEventProcessor(exeCtx utils.ExecutionCtx, batchSize, queueSize int, flushInterval time.Duration) *QueueingEventProcessor {
+	p := NewCustomEventProcessor(exeCtx, batchSize, queueSize, flushInterval, NewInMemoryQueue(queueSize), NewQueueEventDispatcher(exeCtx.GetContext()))
+	p.BatchSize = DefaultBatchSize
+	if batchSize > 0 {
+		p.BatchSize = batchSize
+	}
+
+	p.StartTicker(exeCtx.GetContext())
+	return p
+}
+
+// NewCustomEventProcessor returns a new QueueingEventProcessor with the queue passed in and the dispatcher passed in.
+func NewCustomEventProcessor(exeCtx utils.ExecutionCtx, batchSize, queueSize int, flushInterval time.Duration, q Queue, dispatcher Dispatcher) *QueueingEventProcessor {
 	p := &QueueingEventProcessor{
 		MaxQueueSize:    queueSize,
 		FlushInterval:   flushInterval,
-		Q:               NewInMemoryQueue(queueSize),
-		EventDispatcher: NewQueueEventDispatcher(exeCtx.GetContext()),
+		Q:               q,
+		EventDispatcher: dispatcher,
 		wg: exeCtx.GetWaitSync(),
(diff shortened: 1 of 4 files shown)


mikeng13:
Also please add unit test for it and put in a summary for the PR

thomaszurkan-optimizely:
> There is a pattern than @pawels-optimizely introduced that seems to be very well adopted in the Go community called option funcs shown here: https://github.com/optimizely/go-sdk/blob/master/optimizely/client/factory.go#L100
> 
> Maybe we can start adopting this pattern more.

1. What does this buy us here?
2. Does this mean that we want function chaining?

thomaszurkan-optimizely:
> Also please add unit test for it and put in a summary for the PR

I'm not sure additional unit tests are necessary here because it is used internally in the NewEventProcessor call which is thoroughly tested.

mikeng13:
> > There is a pattern than @pawels-optimizely introduced that seems to be very well adopted in the Go community called option funcs shown here: https://github.com/optimizely/go-sdk/blob/master/optimizely/client/factory.go#L100
> > Maybe we can start adopting this pattern more.
> 
> 1. What does this buy us here?
> 2. Does this mean that we want function chaining?

Not necessarily chaining, but what it buys us is the ability to have variable arguments in the factory method, making it easier to extend in the future if we want to add additional configuration options.

pawels-optimizely:
as I read correctly,  NewEventProcessor starts its own ticker. but it also calls NewCustomEventProcessor which starts its own ticker too. so we have 2 tickers running at the same time. also there is some repetitive logic on line 62-65, which is same as in NewCustomEventProcessor. 

as far as optional parameters go, we can implement that for batchSize, queueSize,  and flushInterval. These are good candidates I think, and then initialization will be simpler: you won't have to have if statements - just initialize with default values and run through optional parameters.

optimizely/event/processor.go
```diff
@@ -56,21 +56,71 @@ const DefaultEventFlushInterval = 30 * time.Second
 
 var pLogger = logging.GetLogger("EventProcessor")
 
+// QPOption is the QueuingProcessor options that give you the ability to add one more more options before the processor is initialized.
+type QPConfigOption func(qp *QueueingEventProcessor)
+
+func ProcessorBatchSize(bsize int) QPConfigOption {
```
mikeng13: nit: Not sure we need to prepend `Processor` to each of these. Seems redundant since the file name is already processor

optimizely/event/processor_test.go
```diff
@@ -83,23 +102,92 @@ func TestDefaultEventProcessor_ProcessBatch(t *testing.T) {
...
+
+	impression := BuildTestImpressionEvent()
+	conversion := BuildTestConversionEvent()
+
+	processor.ProcessEvent(impression)
+	processor.ProcessEvent(impression)
+
+	assert.Equal(t, 2, processor.EventsCount())
+
+	timers.Sleep(exeCtx.GetContext(), 100 * time.Millisecond)
```
mikeng13: Is there a reason you need to use a 3rd party timer? I'd prefer if we can just keep it simple with out-of-the-box. The reason is that we'd have to document yet another library we use in the OSS tracker

optimizely/event/processor.go
thomaszurkan-optimizely: I thought there might be a class with dispatcher.  but, you may be right.

PR #114, opened 2019-09-23
Added project config update notification

pawels-optimizely:


optimizely/config/polling_manager.go
```diff
@@ -37,18 +38,20 @@ var cmLogger = logging.GetLogger("PollingConfigManager")
 
 // PollingProjectConfigManagerOptions used to create an instance with custom configuration
 type PollingProjectConfigManagerOptions struct {
-	Datafile        []byte
-	PollingInterval time.Duration
-	Requester       utils.Requester
+	Datafile           []byte
+	PollingInterval    time.Duration
+	Requester          utils.Requester
+	NotificationCenter notification.Center
```
mikeng13: What are your thoughts on making the `NotificationCenter` a singleton type package, such that we do not need each component to have a handle/reference to it in order to subscribe to the notifications? I realize that is a more heavy refactor, but I think it would make sense from the perspective of not having to worry about this struct during the construction of each of our top-level components. Another con I see with using a singleton is that it would be harder to unit-test properly. What do you think? @thomaszurkan-optimizely would like your thoughts on this as well

optimizely/notification/center.go
```diff
@@ -34,8 +34,10 @@ type DefaultCenter struct {
 // NewNotificationCenter returns a new notification center
 func NewNotificationCenter() *DefaultCenter {
 	decisionNotificationManager := NewAtomicManager()
+	projectConfigUpdateNotificationManager := NewAtomicManager()
```
mikeng13: This should be generic like `projectConfigNotificationManager` as we can later have other notifications aside from the` Update`

PR #115, opened 2019-09-24
fix(datafile-log): Log added for initial datafile assignment.

yasirfolio3:
1. Added log for the scenario when datafile is initially assigned according to current `java` implementation.

### DIFF
--- a/optimizely/config/polling_manager.go
+++ b/optimizely/config/polling_manager.go
@@ -80,7 +80,7 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
 		return
 	}
 
-	previousRevision := "null"
+	var previousRevision string
 	if cm.projectConfig != nil {
 		previousRevision = cm.projectConfig.GetRevision()
 	}
--- a/optimizely/config/polling_manager.go
+++ b/optimizely/config/polling_manager.go
@@ -69,16 +69,20 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
 	}
 
 	projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(datafile)
-	if err != nil {
-		cmLogger.Error("failed to create project config", err)
-	}
 
 	cm.configLock.Lock()
-	if cm.projectConfig != nil {
-		if cm.projectConfig.GetRevision() == projectConfig.GetRevision() {
+	if err != nil {
+		cmLogger.Error("failed to create project config", err)
+	} else {
+		var previousRevision = "null"
+		if cm.projectConfig != nil {
+			previousRevision = cm.projectConfig.GetRevision()
+		}
+
+		if projectConfig.GetRevision() == previousRevision {
 			cmLogger.Debug(fmt.Sprintf("No datafile updates. Current revision number: %s", cm.projectConfig.GetRevision()))
 		} else {
-			cmLogger.Debug(fmt.Sprintf("Received new datafile and updated config. Old revision number: %s. New revision number: %s", cm.projectConfig.GetRevision(), projectConfig.GetRevision()))
+			cmLogger.Debug(fmt.Sprintf("New datafile set with revision: %s. Old revision: %s", projectConfig.GetRevision(), previousRevision))
 			cm.projectConfig = projectConfig
 
 			if cm.notificationCenter != nil {
@@ -91,8 +95,6 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
 				}
 			}
 		}
-	} else {
-		cm.projectConfig = projectConfig
 	}
 	cm.err = err
 	cm.configLock.Unlock()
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,9 +20,9 @@ jobs:
   include:
     - stage: 'Lint'
       install:
-        - go get github.com/golangci/golangci-lint/cmd/golangci-lint
+        - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.19.0
```
mikeng13: What are these changes for? They seem unrelated to this PR. If so, please put it in another PR.

optimizely/config/polling_manager.go
```diff
@@ -69,16 +69,20 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
 	}
 
 	projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(datafile)
+
+	cm.configLock.Lock()
 	if err != nil {
 		cmLogger.Error("failed to create project config", err)
-	}
+	} else {
+		var previousRevision = "null"
```
mikeng13: This is not right. You are trying to set a zero value string right? You should just declare the variable 'var previousRevision string'

.travis.yml
msohailhussain: linter was causing problem, yeah this should be in a separate PR.

optimizely/config/polling_manager.go
yasirfolio3: actually not, just following java implementation, it prints "null" in case of initial datafile assignment, it can be seen here:
https://github.com/optimizely/java-sdk/blob/c3e55cec05a18dc2761ae7c64d98933efdbbcbf4/core-api/src/main/java/com/optimizely/ab/config/PollingProjectConfigManager.java#L93

.travis.yml
yasirfolio3: Have created a separate PR for this linter-fix now.

yasirfolio3:
> let's make our logic cleaner, keeping in mind that you are returning an error, and that err != nil is the same as cm.projectConfig == nil . Checking for both does not seem right and is very confusing.

`err != nil` checks if datafile was properly converted to projectconfig, whereas `cm.projectConfig != nil` checks for whether a project config already existed in `PollingProjectConfigManager`, Am i missing something?

pawels-optimizely:
with this logic: 
```
projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(datafile)
	if err != nil {
		cmLogger.Error("failed to create project config", err)
	}
```
I feel like we need to do something else: probably return right away.  Then the following logic could be cleaner. The only thing that I am not sure of is  whether we still want to assign projectConfig to nil , even when previous  projectConfig was not nil.

yasirfolio3:
> I just talked to @mikeng13 and he said we should not assign nil to our polling mgr, so you can just return with the error message, leaving cm.projectConfig intact

i think it's already part of this PR? the assignment code is in the `else` whereas `if` checks if new project config was properly created or not.
Heres the check for that:
https://github.com/optimizely/go-sdk/blob/d2f9a5c42a2748b582b61e4308811d7aabbd9488/optimizely/config/polling_manager.go#L74

pawels-optimizely:
I meant something like that, please test it

```	
cm.configLock.Lock()
defer func() {
        cm.err = err
	cm.configLock.Unlock()
}()
	if err != nil {
		cmLogger.Error("failed to create project config", err)
                return 
	} 
	
	previousRevision := cm.projectConfig.GetRevision()
	

		if projectConfig.GetRevision() == previousRevision {
			cmLogger.Debug(fmt.Sprintf("No datafile updates. Current revision number: %s", cm.projectConfig.GetRevision()))
                    return 
		} 
		cmLogger.Debug(fmt.Sprintf("New datafile set with revision: %s. Old revision: %s", projectConfig.GetRevision(), previousRevision))
			cm.projectConfig = projectConfig

			if cm.notificationCenter != nil {
				projectConfigUpdateNotification := notification.ProjectConfigUpdateNotification{
					Type:     notification.ProjectConfigUpdate,
					Revision: cm.projectConfig.GetRevision(),
				}
				if err = cm.notificationCenter.Send(notification.ProjectConfigUpdate, projectConfigUpdateNotification); err != nil {
					cmLogger.Warning("Problem with sending notification")
				}
			}
}
```

yasirfolio3:
@pawels-optimizely tried the above way earlier, tried your code too, it crashes with this error 
`panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference`
when `previousRevision := cm.projectConfig.GetRevision()` is called with `cm.projectConfig` being nil.

yasirfolio3:
@pawels-optimizely have made suggested changes with minor updates, can you please review now.

pawels-optimizely:
ok, that means that GetRevision does not handle nil, in that case you fix GetRevision or wrap it in if statement  : var previousRevision string and then if statement

mikeng13:
@yasirfolio3 can you add unit tests for this

yasirfolio3:
> @yasirfolio3 can you add unit tests for this

this PR has just introduced 1 new log, all other changes are just restructuring of code which was already covered with tests in #104 .

PR #116, opened 2019-09-24
refac(notification): Refactor to global notification center

mikeng13:
# Summary
- Introduces a global notification `Registry` for which components can register notification handlers using the SDK Key.
- Components that want to use the notification registry will need to know about the `sdk_key`
- The registry simply maintains a map of `sdk_key` to `NotificationCenter` and will route accordingly when `RegisterHandler, RemoveHandler, Send` APIs are called on it.

# Tests
- Unit tests (WIP)


optimizely/decision/composite_service.go
```diff
@@ -31,15 +31,15 @@ var csLogger = logging.GetLogger("CompositeDecisionService")
 type CompositeService struct {
 	// experimentDecisionServices []ExperimentDecisionService
 	featureDecisionServices []FeatureService
-	notificationCenter      notification.Center
+	sdkKey                  string
 }
 
 // NewCompositeService returns a new instance of the DefeaultDecisionEngine
-func NewCompositeService(notificationCenter notification.Center) *CompositeService {
+func NewCompositeService(sdkKey string) *CompositeService {
```
mikeng13: The one caveat about this impl is that the component needs to know about the SDK Key, which I don't think it might be too much to ask since customers are already providing it to the client so even if they instantiate a new composite service outside the client, they'll still have access to the SDK Key

mikeng13:
This is WIP, putting up for early feedback.

optimizely/notification/register.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Package notification //
+package notification
+
+import (
+	"fmt"
+)
+
+// Maintains mapping of SDK key to notification center
+var notificationCenterRegistry = make(map[string]Center)
```
thomaszurkan-optimizely: This should be wrapped with a mutex for access.

optimizely/decision/composite_service_test.go
```diff
@@ -76,10 +76,9 @@ func TestOnDecision(t *testing.T) {
 	testFeatureDecisionService := new(MockFeatureDecisionService)
 	testFeatureDecisionService.On("GetDecision", decisionContext, userContext).Return(expectedFeatureDecision, nil)
 
-	notificationCenter := notification.NewNotificationCenter()
 	decisionService := &CompositeService{
```
thomaszurkan-optimizely: Does the composite service need to know about the SDK key? Or, can the registry take care of that?

mikeng13:
Closing this in favor of: https://github.com/optimizely/go-sdk/pull/120

PR #117, opened 2019-09-24
fix(linter): Fixed golangci-lint implementations

yasirfolio3:
1. Fixed implementation of golangci-lint as suggested in the documentation. The previously implemented way was only suggested for local installation and was flaky.
2. Made a minor fix as per linter recommendation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,9 @@ jobs:
   include:
     - stage: 'Lint'
       install:
-        - go get github.com/golangci/golangci-lint/cmd/golangci-lint
+        - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.19.0
       script:
-        - golangci-lint run --out-format=tab --tests=false optimizely/...
+        - $GOPATH/bin/golangci-lint run --out-format=tab --tests=false optimizely/...
     - stage: 'Integration Tests'
       merge_mode: replace
       env: SDK=go
@@ -34,4 +34,3 @@ jobs:
       script:
         - "$HOME/travisci-tools/fsc-trigger/trigger_fullstack-sdk-compat.sh"
       after_success: travis_terminate 0
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,3 +34,4 @@ jobs:
       script:
         - "$HOME/travisci-tools/fsc-trigger/trigger_fullstack-sdk-compat.sh"
       after_success: travis_terminate 0
+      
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -33,5 +33,4 @@ jobs:
         - mkdir $HOME/travisci-tools && pushd $HOME/travisci-tools && git init && git pull https://$CI_USER_TOKEN@github.com/optimizely/travisci-tools.git && popd
       script:
         - "$HOME/travisci-tools/fsc-trigger/trigger_fullstack-sdk-compat.sh"
-      after_success: travis_terminate 0
-
+      after_success: travis_terminate 0
```
pawels-optimizely: no change needed here, new line is probably nicer

optimizely/notification/manager.go
```diff
@@ -51,9 +51,7 @@ func (am *AtomicManager) Add(newHandler func(interface{})) (int, error) {
 // Remove removes handler with the given id
 func (am *AtomicManager) Remove(id int) {
 	handlerID := uint32(id)
-	if _, ok := am.handlers[handlerID]; ok {
```
pawels-optimizely: can we keep that and log a message if the id is not in the map ? just return within the "if" statement, and then log after it.

yasirfolio3: delete works even if key doesn't exist because of which linter was throwing error while using the above condition. We can check for !ok, log and return, and then call delete on the dictionary?

pawels-optimizely: hmm, I know, but my point was to send the warning in case the id did not exist and you try to delete something that did not exist - to me that scenario should not happen. I guess that's fine then if linters complain. No worries.

yasirfolio3: > hmm, I know, but my point was to send the warning in case the id did not exist and you try to delete something that did not exist - to me that scenario should not happen. I guess that's fine then if linters complain. No worries.

yes i got your point, just bound by the linter warning, thanks.

pawels-optimizely: can you try this: 
```
if _, ok := am.handlers[handlerID]; ok {
		delete(am.handlers, handlerID)
                 return
	}
        // log here 
```

yasirfolio3: `return` statement seems to have done the trick for avoiding linter warning, thanks.

pawels-optimizely: great, smart  linters :), they figured out that nothing was done when item not in a map.

.travis.yml
```diff
@@ -34,4 +34,4 @@ jobs:
       script:
         - "$HOME/travisci-tools/fsc-trigger/trigger_fullstack-sdk-compat.sh"
       after_success: travis_terminate 0
-
+
```
mikeng13: Please add a newline here

PR #119, opened 2019-09-25
refactor initialization part  of optimizely client and polling manager

pawels-optimizely:


optimizely/client/factory_test.go
```diff
@@ -48,41 +49,23 @@ func TestFactoryClientReturnsDefaultClient(t *testing.T) {
 func TestClientWithSDKKey(t *testing.T) {
 	factory := OptimizelyFactory{SDKKey: "1212"}
 
-	clientOptions := Options{}
-
-	client, err := factory.ClientWithOptions(clientOptions)
+	app, err := factory.Client()
```
mikeng13: Not: instead of app it can just be optimizelyClient. I think that is a more accurate name for it.

optimizely/config/polling_manager.go
```diff
@@ -29,31 +29,67 @@ import (
...
+	err           error
+	projectConfig optimizely.ProjectConfig
+	exeCtx        utils.ExecutionCtx // context used for execution control
+}
+
+// OptionFunc is a type to a proper func
+type OptionFunc func(*PollingProjectConfigManager)
+
+// SetDefaultRequester is an optional function, sets default requester based on a key.
+func SetDefaultRequester(sdkKey string) OptionFunc {
```
mikeng13: It's inconsistent that the optionfunc for the polling manager starts with "Set" while the ones in the optly client don't. Let's get to keep them consistent.

optimizely/config/polling_manager.go
```diff
@@ -64,29 +64,29 @@ func SetDefaultRequester(sdkKey string) OptionFunc {
...
+// PollingInterval is an optional function, sets a passed polling interval
+func PollingInterval(interval time.Duration) OptionFunc {
 	return func(p *PollingProjectConfigManager) {
 		p.pollingInterval = interval
 	}
 }
 
-// SetNotification is an optional function, sets a passed notification
-func SetNotification(notificationCenter notification.Center) OptionFunc {
+// Notification is an optional function, sets a passed notification
```
mikeng13: nit: This should be `NotificationCenter`

pawels-optimizely: changed

optimizely/config/polling_manager.go
pawels-optimizely: done

optimizely/client/factory_test.go
pawels-optimizely: done

PR #120, opened 2019-09-26
feat(registry): Added service registry for retrieving instances of notif center

mikeng13:
# Summary
- Refactors how we construct top-level components with the notification center
- This simplifies the initialization code for PollingConfigManager and DecisionService because now they don't need to know about the internal notification center

# Tests
- Unit tests (missing) for the service registry


mikeng13:
> Looks good. Are you sure you want to make the registry methods public?

Is there a way to make it "protected" so that only our packages can access it?

Issue #59, opened 2019-08-13
Activate and GetVariation roadmap


etiennemarais:
### Question

What does the roadmap look for the go-sdk in terms of feature parity? At the moment only `IsFeatureEnabled` is supported.

mikeng13:
Thanks for reaching out! We have plans to bring the Go SDK to parity with the rest of the SDKs in the coming months. We don't have an exact date yet but we are prioritizing bringing feature experimentation to the Go SDK first before adding Activate and Track. 

Hope this helps!

I will leave this open until we've delivered.

mikeng13:
This is now available in https://github.com/optimizely/go-sdk/releases/tag/v1.0.0-beta2

Note that there are still some missing features because we're in beta: whitelisting, forced variations, user profile service.

Please take a look! We would appreciate any feedback!

PR #122, opened 2019-09-28
feat(activate): Add Activate and GetVariation APIs.

mikeng13:
# Summary
- Add `Activate` API for A/B tests
- Add `GetVariation` API for A/B tests
- Introduced test suites and test fixtures for client package testing
- Adds Decision notifications for both

# Tests
- Unit

Address #59

### DIFF
--- a/optimizely/client/client.go
+++ b/optimizely/client/client.go
@@ -42,6 +42,41 @@ type OptimizelyClient struct {
 	executionCtx utils.ExecutionCtx
 }
 
+// Activate returns the key of the variation the user is bucketed into and sends an impression event to the Optimizely log endpoint
+func (o *OptimizelyClient) Activate(experimentKey string, userContext entities.UserContext) (result string, err error) {
+
+	defer func() {
+		if r := recover(); r != nil {
+			switch t := r.(type) {
+			case error:
+				err = t
+			case string:
+				err = errors.New(t)
+			default:
+				err = errors.New("unexpected error")
+			}
+			errorMessage := fmt.Sprintf("optimizely SDK is panicking with the error:")
+			logger.Error(errorMessage, err)
+			logger.Debug(string(debug.Stack()))
+		}
+	}()
+
+	decisionContext, experimentDecision, err := o.getExperimentDecision(experimentKey, userContext)
+	if err != nil {
+		logger.Error("received an error while computing experiment decision", err)
+		return result, err
+	}
+
+	if experimentDecision.Variation != nil {
+		// send an impression event
+		result = experimentDecision.Variation.Key
+		impressionEvent := event.CreateImpressionUserEvent(decisionContext.ProjectConfig, *decisionContext.Experiment, *experimentDecision.Variation, userContext)
+		o.eventProcessor.ProcessEvent(impressionEvent)
+	}
+
+	return result, err
+}
+
 // IsFeatureEnabled returns true if the feature is enabled for the given user
 func (o *OptimizelyClient) IsFeatureEnabled(featureKey string, userContext entities.UserContext) (result bool, err error) {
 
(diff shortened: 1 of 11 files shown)


optimizely/client/client.go
```diff
@@ -42,6 +42,41 @@ type OptimizelyClient struct {
...
+		}
+	}()
+
+	decisionContext, experimentDecision, err := o.getExperimentDecision(experimentKey, userContext)
+	if err != nil {
+		logger.Error("received an error while computing experiment decision", err)
+		return result, err
+	}
+
+	if experimentDecision.Variation != nil {
```
pawels-optimizely: can you have nil error and variation still nil ?

optimizely/client/client.go
```diff
@@ -364,43 +360,107 @@ func (o *OptimizelyClient) GetAllFeatureVariables(featureKey string, userContext
...
+			logger.Debug(string(debug.Stack()))
+		}
+	}()
+
+	_, experimentDecision, err := o.getExperimentDecision(experimentKey, userContext)
+	if err != nil {
+		logger.Error("received an error while computing experiment decision", err)
+	}
+
+	if experimentDecision.Variation != nil {
```
pawels-optimizely: same as previous

optimizely/client/client.go
```diff
@@ -364,43 +360,107 @@ func (o *OptimizelyClient) GetAllFeatureVariables(featureKey string, userContext
...
+
+	projectConfig, err := o.GetProjectConfig()
+	if err != nil {
+		logger.Error("Optimizely SDK tracking error", err)
+		return err
+	}
+
+	configEvent, err := projectConfig.GetEventByKey(eventKey)
+
+	if err == nil {
```
pawels-optimizely: if possible, check for not nil error first  - it will eliminate else/if - and it will be cleaner

optimizely/decision/composite_service.go
```diff
@@ -78,6 +78,30 @@ func (s CompositeService) GetFeatureDecision(featureDecisionContext FeatureDecis
 	return featureDecision, err
 }
 
+// GetExperimentDecision returns a decision for the given experiment key
+func (s CompositeService) GetExperimentDecision(experimentDecisionContext ExperimentDecisionContext, userContext entities.UserContext) (experimentDecision ExperimentDecision, err error) {
+	experimentDecision, err = s.compositeExperimentService.GetDecision(experimentDecisionContext, userContext)
```
pawels-optimizely: not checking for err here and possibly assigning err again on line 97

optimizely/client/client.go
```diff
@@ -364,43 +360,107 @@ func (o *OptimizelyClient) GetAllFeatureVariables(featureKey string, userContext
...
 	}()
 
 	userID := userContext.ID
 	logger.Debug(fmt.Sprintf(`Evaluating feature "%s" for user "%s".`, featureKey, userID))
 
-	projectConfig, err := o.GetProjectConfig()
-	if err != nil {
-		logger.Error("Error calling getFeatureDecision", err)
-		return decisionContext, featureDecision, err
+	projectConfig, e := o.GetProjectConfig()
```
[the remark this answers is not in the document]
mikeng13: Because we already named the return error parameter `err` and we don't want to re-assign it here. Here is an article that explains it: https://yourbasic.org/golang/gotcha-shadowing-variables/

optimizely/client/client.go
```diff
@@ -415,10 +475,48 @@ func (o *OptimizelyClient) getFeatureDecision(featureKey string, userContext ent
 		return decisionContext, featureDecision, err
 	}
 
-	// @TODO(yasir): send decision notification
 	return decisionContext, featureDecision, err
 }
 
+func (o *OptimizelyClient) getExperimentDecision(experimentKey string, userContext entities.UserContext) (decisionContext decision.ExperimentDecisionContext, experimentDecision decision.ExperimentDecision, err error) {
+
+	userID := userContext.ID
```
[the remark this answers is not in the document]
mikeng13: I think we allow empty string user IDs. In the other SDKs we simply check that they can't be `nil`, but here in Go strings cannot be `nil`

optimizely/client/client.go
```diff
@@ -415,10 +475,48 @@ func (o *OptimizelyClient) getFeatureDecision(featureKey string, userContext ent
...
+	}
+
+	decisionContext = decision.ExperimentDecisionContext{
+		Experiment:    &experiment,
+		ProjectConfig: projectConfig,
+	}
+
+	experimentDecision, err = o.decisionService.GetExperimentDecision(decisionContext, userContext)
+	if err != nil {
+		err = nil
```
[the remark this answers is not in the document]
mikeng13: You are right. This shouldn't be set to nil.

optimizely/client/client.go
mikeng13: Yep, what @mjc1283 said

optimizely/client/client.go
mikeng13: Yeah same response. You can have both nil variation and nil error

optimizely/client/client.go
mikeng13: ok

mikeng13:
Address #59

PR #124, opened 2019-10-01
Add ExperimentWhitelistService

mjc1283:


optimizely/decision/experiment_whitelist_service.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Package decision //
+package decision
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/optimizely/go-sdk/optimizely/decision/reasons"
+
```
pawels-optimizely: no new line, same group

optimizely/decision/experiment_whitelist_service.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// ExperimentWhitelistService makes a decision using a whitelist (a set of experiment + variation assignments for a set of users)
+// whitelist should be a map of user ID, to a map of Experiment key to Variation key
+type ExperimentWhitelistService struct {
+	whitelist map[string]map[string]string
+}
+
+// NewExperimentWhitelistService returns a new instance of ExperimentWhitelistService
+func NewExperimentWhitelistService(whitelist map[string]map[string]string) *ExperimentWhitelistService {
```
pawels-optimizely: who/what will create whitelist map ?

optimizely/decision/experiment_whitelist_service.go
```diff
@@ -0,0 +1,74 @@
...
+
+// GetDecision returns a decision with a variation when an entry is found for a given user ID and experiment key
+func (s ExperimentWhitelistService) GetDecision(decisionContext ExperimentDecisionContext, userContext entities.UserContext) (decision ExperimentDecision, err error) {
+	if decisionContext.Experiment == nil {
+		return decision, errors.New("decisionContext Experiment is nil")
+	}
+
+	experiment, err := decisionContext.ProjectConfig.GetExperimentByKey(decisionContext.Experiment.Key)
+	if err != nil {
+		return decision, fmt.Errorf("error looking up experiment in decision context: %w", err)
```
pawels-optimizely: have not seen %w, you mean %v?

optimizely/decision/experiment_whitelist_service.go
mjc1283: We could have the factory option accept this map and pass it to this service. But do you think there's a better/more convenient format for the user use to specify the whitelist?

optimizely/decision/experiment_whitelist_service.go
mjc1283: I think this allows `fmt.Errorf` to return a wrapped error.

>If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand. It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. The %w verb is otherwise a synonym for %v.

(from https://golang.org/pkg/fmt/#Errorf)

But this is not supported in older Go versions, so I will change it to `%v`.

PR #126, opened 2019-10-01
chore: Prepare release 1.0.0-beta2.

mikeng13:
# Summary
- Bumps to 1.0.0-beta2
- Adds changelog entry

### DIFF
--- a/optimizely/version.go
+++ b/optimizely/version.go
@@ -18,7 +18,7 @@
 package optimizely
 
 // Version is the current version of the client
-const Version = "0.2.0"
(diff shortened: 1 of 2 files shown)


optimizely/version.go
```diff
@@ -18,7 +18,7 @@
 package optimizely
 
 // Version is the current version of the client
-const Version = "0.2.0"
```
aliabbasrizvi: So, we published `1.0.0-beta1` with this version?

mikeng13: Yes this was not bumped

PR #128, opened 2019-10-02
feat(notif): Add notification handlers from factory methods.

mikeng13:
# Summary
- Introduce option func for adding notification callbacks in the DecisionService and ConfigManager
- Introduces a new option func paradigm where the func name is prepended with `With` to avoid naming collisions as well as making it more apparent that they are option func methods.

# Tests
- Unit


mikeng13:
Not using this as it doesn't give us much additional value

PR #129, opened 2019-10-02
feat: Add ExperimentWhitelistService

mjc1283:
## Summary

Add `ExperimentWhitelistService`, which returns decisions based on the contents of `forcedVariations` in experiments in the datafile. `NewCompositeExperimentService` creates an `ExperimentWhitelistService` as its first child service, taking precedence over normal bucketing.

To support `ExperimentWhitelistService`, I added a `Whitelist` field to the `Experiment` entity. The value comes directly from `forcedVariations` of the datafile experiment. I named it `Whitelist` to match the term we use to talk about this kind of override, even though it's not named that in the datafile.

## Test plan
Added unit tests

### DIFF
--- a/optimizely/decision/composite_experiment_service.go
+++ b/optimizely/decision/composite_experiment_service.go
@@ -29,10 +29,12 @@ type CompositeExperimentService struct {
 // NewCompositeExperimentService creates a new instance of the CompositeExperimentService
 func NewCompositeExperimentService() *CompositeExperimentService {
 	// These decision services are applied in order:
-	// 1. Bucketing
-	// @TODO(mng): Prepend forced variation and whitelisting services
+	// 1. Whitelist
+	// 2. Bucketing
+	// @TODO(mng): Prepend forced variation
 	return &CompositeExperimentService{
 		experimentServices: []ExperimentService{
+			NewExperimentWhitelistService(),
 			NewExperimentBucketerService(),
 		},
 	}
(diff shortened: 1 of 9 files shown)


optimizely/decision/composite_experiment_service.go
```diff
@@ -29,10 +29,12 @@ type CompositeExperimentService struct {
...
 func NewCompositeExperimentService() *CompositeExperimentService {
 	// These decision services are applied in order:
-	// 1. Bucketing
-	// @TODO(mng): Prepend forced variation and whitelisting services
+	// 1. Whitelist
+	// 2. Bucketing
+	// @TODO(mng): Prepend forced variation
 	return &CompositeExperimentService{
 		experimentServices: []ExperimentService{
+			NewExperimentWhitelistService(),
```
mikeng13: I love the simplicity of this 😄

optimizely/decision/experiment_whitelist_service.go
```diff
@@ -0,0 +1,58 @@
...
+package decision
+
+import (
+	"errors"
+
+	"github.com/optimizely/go-sdk/optimizely/decision/reasons"
+	"github.com/optimizely/go-sdk/optimizely/entities"
+)
+
+// ExperimentWhitelistService makes a decision using an experiment's whitelist (a map of user id to variation keys)
```
mikeng13: I don't think I did this in order classes, but would be good to mention in the docstring that it implements the `ExperimentService` interface

PR #133, opened 2019-10-03
ci(travis): Used updated travis script with gosdkonly

msohailhussain:


msohailhussain:
#129 already have changes

PR #138, opened 2019-10-04
fix(composite services): Don't use decision returned from inner service when it returns an error

mjc1283:
Summary:

- In `CompositeExperimentService` and `CompositeFeatureService` `GetDecision` methods, when the inner service returns an error, skip that decision and move to the next inner service
- Return an error from `CompositeExperimentService` `GetDecision` when no decision was made

Test Plan:

Updated & added unit tests


pkg/decision/composite_experiment_service.go
```diff
@@ -47,10 +53,14 @@ func (s CompositeExperimentService) GetDecision(decisionContext ExperimentDecisi
...
+		if err != nil {
+			ceLogger.Debug(fmt.Sprintf("%v", err))
+		}
+		if decision.Variation != nil && err == nil {
 			return decision, err
 		}
 	}
 
-	return experimentDecision, nil
+	return experimentDecision, errors.New("no decision was made")
```
mikeng13: Should we log this error or should the parent `composite_service` log it?

pkg/decision/composite_feature_service.go
```diff
@@ -45,10 +46,13 @@ func NewCompositeFeatureService(compositeExperimentService ExperimentService) *C
...
+		if err != nil {
+			cfLogger.Debug(fmt.Sprintf("%v", err))
+		}
+		if featureDecision.Variation != nil && err == nil {
 			return featureDecision, err
 		}
 	}
 
-	return FeatureDecision{}, fmt.Errorf("no decision was made for feature %s", decisionContext.Feature.Key)
+	return FeatureDecision{}, errors.New("no decision was made")
```
mikeng13: Same thing re: logging

pkg/decision/composite_experiment_service.go
mjc1283: If I'm reading correctly, the client will end up logging this error, and the same is true for the composite feature service.

mikeng13: Ah got it, ship away then!

PR #140, opened 2019-10-08
(feat): first pass at using sdk key with processor for notification events.  …

thomaszurkan-optimizely:
…this should be refactored to use the sdk key in the project config.

### DIFF
--- a/pkg/event/processor.go
+++ b/pkg/event/processor.go
@@ -18,13 +18,16 @@
 package event
 
 import (
+	"bytes"
+	"encoding/json"
 	"errors"
+	"fmt"
+	"github.com/optimizely/go-sdk/pkg/logging"
+	"github.com/optimizely/go-sdk/pkg/notification"
+	"github.com/optimizely/go-sdk/pkg/registry"
+	"github.com/optimizely/go-sdk/pkg/utils"
 	"sync"
 	"time"
-
-	"github.com/optimizely/go-sdk/pkg/utils"
-
-	"github.com/optimizely/go-sdk/pkg/logging"
 )
 
 // Processor processes events
@@ -34,6 +37,7 @@ type Processor interface {
 
 // QueueingEventProcessor is used out of the box by the SDK
 type QueueingEventProcessor struct {
+	sdkKey 			string
 	MaxQueueSize    int           // max size of the queue before flush
 	FlushInterval   time.Duration // in milliseconds
 	BatchSize       int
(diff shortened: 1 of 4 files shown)


pkg/event/processor.go
```diff
@@ -18,13 +18,16 @@
 package event
 
 import (
+	"bytes"
```
pawels-optimizely: nit: group the packages properly

pkg/event/processor.go
```diff
@@ -92,6 +96,13 @@ func PDispatcher(d Dispatcher) QPConfigOption {
 	}
 }
 
+// PDispatcher sets the Processor Dispatcher as a config option to be passed into the NewProcessor method
```
pawels-optimizely: bad comment - linters will complain

pkg/event/processor.go
```diff
@@ -210,7 +221,7 @@ func (p *QueueingEventProcessor) FlushEvents() {
 
 	for p.EventsCount() > 0 {
 		if failedToSend {
-			pLogger.Error("last Event Batch failed to send; retry on next flush", errors.New("dispatcher failed"))
+			pLogger.Error("last Event Batch failed to send; retry on next flush", errors.New("Dispatcher failed"))
```
pawels-optimizely: errors usually don't start with an upper case letter - linters should complain

pkg/event/processor.go
```diff
@@ -242,7 +253,27 @@ func (p *QueueingEventProcessor) FlushEvents() {
 		}
 		if batchEventCount > 0 {
 			// TODO: figure out what to do with the error
-			if success, _ := p.EventDispatcher.DispatchEvent(createLogEvent(batchEvent)); success {
+			logEvent := createLogEvent(batchEvent)
+			// we have to marshal the log event in order to send.  maybe we should just pass marshaled data to dispatch
+			{
```
pawels-optimizely: do we need {} if no if statement ?

pkg/notification/entities.go
```diff
@@ -49,3 +54,9 @@ type ProjectConfigUpdateNotification struct {
 	Type     Type
 	Revision string
 }
+
+type LogEventNotification struct {
+	Type         Type
+	LogEvent  	 *bytes.Buffer
```
pawels-optimizely: I think it may be better not to convert to bytes , and just pass LogEvent, then Marshal is not needed.

thomaszurkan-optimizely: cyclical import between event and notification packages.

pkg/event/processor.go
```diff
@@ -242,7 +252,15 @@ func (p *QueueingEventProcessor) FlushEvents() {
 		}
 		if batchEventCount > 0 {
 			// TODO: figure out what to do with the error
```
mikeng13: nit: This should go right before we call `DispatchEvent`

pkg/event/processor.go
```diff
@@ -255,3 +273,33 @@ func (p *QueueingEventProcessor) FlushEvents() {
 	}
 	p.Mux.Unlock()
 }
+
+// OnEventDispatch registers a handler for LogEvent notifications
+func (p *QueueingEventProcessor) OnEventDispatch(callback func(eventNotification LogEvent)) (int, error) {
```
mikeng13: nit: param name should just be `logEvent` instead of `eventNotification`

pkg/event/processor.go
thomaszurkan-optimizely: or be greater than or equal which i prefer

PR #141, opened 2019-10-10
refac(ep): Rename options and EP implementation.

mikeng13:
# Summary 
- Rename option func to be prepended with the word "With"
- Rename the QueueingEventProcessor to BatchEventProcessor to be more in line with our other SDKs and the dev docs

### DIFF
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -45,9 +45,9 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 	appClient := &OptimizelyClient{
 		executionCtx:    executionCtx,
 		DecisionService: decision.NewCompositeService(f.SDKKey),
-		EventProcessor: event.NewEventProcessor(event.BatchSize(event.DefaultBatchSize),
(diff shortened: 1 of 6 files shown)


pkg/client/factory.go
```diff
@@ -45,9 +45,9 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 	appClient := &OptimizelyClient{
 		executionCtx:    executionCtx,
 		DecisionService: decision.NewCompositeService(f.SDKKey),
-		EventProcessor: event.NewEventProcessor(event.BatchSize(event.DefaultBatchSize),
-			event.QueueSize(event.DefaultEventQueueSize), event.FlushInterval(event.DefaultEventFlushInterval),
-			event.SDKKey(f.SDKKey)),
+		EventProcessor: event.NewEventProcessor(event.WithBatchSize(event.DefaultBatchSize),
```
msohailhussain: @mikeng13 shouldn't we call it `NewBatchEventProcessor`?

mikeng13: Yeah that is correct. Let me make that change!

PR #145, opened 2019-10-14
(fix): update for tests to run 5 times faster

thomaszurkan-optimizely:


pkg/event/dispatcher.go
```diff
@@ -29,7 +29,7 @@ import (
 
 const maxRetries = 3
 const defaultQueueSize = 1000
-const sleepTime = 5 * time.Second
+const sleepTime = 1 * time.Second
```
mikeng13: At some point we should make this an exponential back-off no?

PR #147, opened 2019-10-18
Expand LogConsumer interface to have explicit field mapping.

mikecdavis:
## Summary
* Updates the `LogConsumer` interface to accept a `fields map[string]string`
* Moves message formatting into the the `LogConsumer` implementation
* Parameterize `io.Writer` in the default `LogConsumer` to facilitate testing
* Updates tests to accommodate changes

I've found the current `LogConsumer` interface a bit too rigid. We have structured elements that we're supporting (i.e. level and name) that are embedded into the message making it very difficult to implement any custom formatting. This change keeps the existing default log format, but allows custom `LogConsumer` implementation to use the supplied `fields` in a matter that best suites in their use case.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,15 +3,12 @@ module github.com/optimizely/go-sdk
 go 1.12
 
 require (
-	github.com/DATA-DOG/godog v0.7.13
 	github.com/google/uuid v1.1.1
 	github.com/json-iterator/go v1.1.7
-	github.com/optimizely/subset v0.0.0-20191016231344-a0cb13c639ec
 	github.com/pkg/profile v1.3.0
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/testify v1.4.0
 	github.com/twmb/murmur3 v1.0.0
-	gopkg.in/yaml.v3 v3.0.0-20191010095647-fc94e3f71652
 )
 
 // Work around issue wtih git.apache.org/thrift.git
--- a/go.mod
+++ b/go.mod
@@ -3,12 +3,15 @@ module github.com/optimizely/go-sdk
 go 1.12
 
 require (
+	github.com/DATA-DOG/godog v0.7.13
 	github.com/google/uuid v1.1.1
 	github.com/json-iterator/go v1.1.7
+	github.com/optimizely/subset v0.0.0-20191016231344-a0cb13c639ec
 	github.com/pkg/profile v1.3.0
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -3,12 +3,15 @@ module github.com/optimizely/go-sdk
 go 1.12
 
 require (
+	github.com/DATA-DOG/godog v0.7.13
```
mikeng13: Hmm we need to find a way to ignore these test dependencies when we `go get` 

These are only used for the cucumber step definitions and we are planning on just running in docker to not pollute this go.mod file. I would rather not have to declare test deps in here.

For now the best thing to do is just revert these changes until we figure out a better way to handle test dependencies.

pkg/logging/logger.go
```diff
@@ -59,13 +62,13 @@ func SetLogLevel(logLevel LogLevel) {
 // GetLogger returns a log producer with the given name
 func GetLogger(name string) OptimizelyLogProducer {
 	return NamedLogProducer{
-		name: name,
+		fields: map[string]interface{}{"name": name},
```
mikecdavis: Open to alternative suggestions to `"name"` as the key here. 

Perhaps `"package"` so we wouldn't use the `[Optimizely]` tag that we have in the default consumer, but instead use the actual package that is logging: (e.g. `"optimizely/go-sdk/pkg/client"`, `"optimizely/go-sdk/pkg/config/datafileprojectconfig"`, etc.)

Currently we don't use names that directly map to the struct or package. For example:
```go
var cmLogger = logging.GetLogger("PollingConfigManager")
```
The above name doesn't exist anywhere, only in logs.

Which in this proposal would be:
```go
var logger = logging.GetLogger("optimizely/go-sdk/pkg/config") 
```
since this is the package name that is logging.

This could simplify things as each package we'd have a single logger, but if this is not granular enough we can add the file name.
```go
var cmLogger = logging.GetLogger("optimizely/go-sdk/pkg/config/polling_manager") 
```

mikeng13:
Thanks! I've created a PR here to ignore the integration test mods: https://github.com/optimizely/go-sdk/pull/148

pkg/logging/logger.go
mikeng13: I like this! It's way easier to map where the log messages came from.

mikecdavis: Cool, I'll create a follow-up.

PR #149, opened 2019-10-22
(fix): Fixed GetDecision for CompositeFeatureService and key for VisitorAttribute.

yasirfolio3:
### Summary
- Fixed `GetDecision` method for CompositeFeatureService, it now returns the last decision made in case every decision-service returned an error.
- Fixed assignment of `Key` property for visitor attribute struct.

### DIFF
--- a/pkg/decision/composite_feature_service.go
+++ b/pkg/decision/composite_feature_service.go
@@ -44,19 +44,16 @@ func NewCompositeFeatureService(compositeExperimentService ExperimentService) *C
 
 // GetDecision returns a decision for the given feature and user context
 func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionContext, userContext entities.UserContext) (FeatureDecision, error) {
-	lastElementIndex := len(f.featureServices) - 1
-	for index, featureDecisionService := range f.featureServices {
-		featureDecision, err := featureDecisionService.GetDecision(decisionContext, userContext)
+	var featureDecision = FeatureDecision{}
+	var err = errors.New("")
+	for _, featureDecisionService := range f.featureServices {
+		featureDecision, err = featureDecisionService.GetDecision(decisionContext, userContext)
 		if err != nil {
 			cfLogger.Debug(fmt.Sprintf("%v", err))
 		}
 		if featureDecision.Variation != nil && err == nil {
 			return featureDecision, err
 		}
-		if index == lastElementIndex {
-			return featureDecision, err
-		}
 	}
-
-	return FeatureDecision{}, errors.New("no decision was made")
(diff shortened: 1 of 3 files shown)


pkg/decision/composite_feature_service.go
```diff
@@ -44,14 +44,18 @@ func NewCompositeFeatureService(compositeExperimentService ExperimentService) *C
 
 // GetDecision returns a decision for the given feature and user context
 func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionContext, userContext entities.UserContext) (FeatureDecision, error) {
-	for _, featureDecisionService := range f.featureServices {
+	lastElementIndex := len(f.featureServices) - 1
```
mikeng13: There is a cleaner way to do this where you declare a FeatureDecision and an error var. You can then loop through and assign the result of calling `.GetDecision` to it and after the loop you can return whatever those variables were set to

pawels-optimizely: @mikeng13 , I think we need to return err all the time here. "no decision was made" error is not applicable anymore (we can log the warning), but our return statement will be always ```return FeatureDecision{}, err ``` ( err being the last iteration from the loop). is that correct ?

mikeng13: Yes that is correct!

yasirfolio3: @mikeng13  Requested changes made.

pkg/decision/composite_feature_service.go
```diff
@@ -44,19 +44,16 @@ func NewCompositeFeatureService(compositeExperimentService ExperimentService) *C
 
 // GetDecision returns a decision for the given feature and user context
 func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionContext, userContext entities.UserContext) (FeatureDecision, error) {
-	lastElementIndex := len(f.featureServices) - 1
-	for index, featureDecisionService := range f.featureServices {
-		featureDecision, err := featureDecisionService.GetDecision(decisionContext, userContext)
+	var featureDecision = FeatureDecision{}
+	var err = errors.New("")
```
mikeng13: This should just be `var err error` so that it is initialized to nil. Otherwise we always return a non-nil error

yasirfolio3: This was on purpose because if there were no feature services, `var err error` would return nil error, are we ok with it?

mikeng13: Yeah that is fine

yasirfolio3: done, thanks 👍

PR #151, opened 2019-10-23
refac(client): Revise how the client handles and returns errors.

mikeng13:
# Summary

This PR revises how the top-level client handles and returns errors:
- Only fatal errors are returned: invalid/no config and unhandled panics
- Other "errors" are logged as warning and will not necessarily return zero-value result

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -1555,6 +1555,22 @@ func (s *ClientTestSuiteAB) TestActivatePanics() {
 	s.EqualError(err, "I'm panicking")
 }
 
+func (s *ClientTestSuiteAB) TestActivateInvalidConfig() {
+	testUserContext := entities.UserContext{}
+
+	mockConfigManager := new(MockProjectConfigManager)
+	expectedError := errors.New("no project config available")
+	mockConfigManager.On("GetConfig").Return(s.mockConfig, expectedError)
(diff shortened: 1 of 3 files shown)


pkg/client/client_test.go
```diff
@@ -1719,8 +1723,10 @@ func (s *ClientTestSuiteFM) TestIsFeatureEnabledWithDecisionError() {
 		ConfigManager:   s.mockConfigManager,
 		DecisionService: s.mockDecisionService,
 	}
+
+	// should still return the decision because the error is non-fatal
 	result, err := client.IsFeatureEnabled(testFeature.Key, testUserContext)
-	s.False(result)
+	s.True(result)
 	s.NotNil(err)
```
mikecdavis: We're returning an `err` _and_ a decision? I thought we wanted to swallow these errors if they're not fatal. Was this this missed or intentionally kept?

Also we should change to either `s.Error(err)` or s`.NoError(err)` for consistency :)

mikeng13: I missed this. We shouldn't return the error. Thanks for catching!

pawels-optimizely: the only true error returned is from GetProjectConfig() method;  getFeatureDecision() has that call (indirectly). if we have an error from GetProjectConfig() then I still think we need to return it - that's a legit error.

PR #152, opened 2019-10-25
feat(gherkin): Integration of gherkin tests as stage in travis.

yasirfolio3:


.travis.yml
```diff
@@ -8,6 +8,7 @@ stages:
   - 'Lint'
   # - 'Integration tests'
   - 'Unit test'
+  - 'Integration as unit tests'
```
mikeng13: This should just be `Integration tests` and you can remove the stage above

.travis.yml
```diff
@@ -54,3 +55,11 @@ jobs:
         - go get github.com/mattn/goveralls
       after_success:
         - $GOPATH/bin/goveralls -coverprofile=profile.cov -service=travis-ci
+    - stage: 'Integration as unit tests'
```
mikeng13: Integration tests

scripts/get-fsc-tests.sh
```diff
@@ -0,0 +1,15 @@
...
+FSC_PATH=tmp/fsc-repo
+rm -rf $FSC_PATH
+mkdir -p $FSC_PATH
+ 
+pushd $FSC_PATH && git init && git fetch --depth=1 https://$CI_USER_TOKEN@github.com/optimizely/fullstack-sdk-compatibility-suite ${FSC_BRANCH:-master} && git checkout FETCH_HEAD && popd
+mkdir -p ./tests/integration/features
+cp -r ./$FSC_PATH/features/* ./tests/integration/features
+mkdir -p ./fsc-datafiles
+cp -r ./$FSC_PATH/features/support/datafiles/*.json ./fsc-datafiles
+export DATAFILES_DIR=$TRAVIS_BUILD_DIR/fsc-datafiles
```
mikeng13: This is specific to just Travis, can it be done in a way that is agnostic or do we need to know about the `TRAVIS_BUILD_DIR` variable? Maybe we can set a generic var in the travis file that this references

mikeng13:
@yasirfolio3 @msohailhussain can we also include tag support so that we run only the tests that are passing and are able to get a green build?

.travis.yml
```diff
@@ -54,3 +54,13 @@ jobs:
...
         - $GOPATH/bin/goveralls -coverprofile=profile.cov -service=travis-ci
+    - stage: 'Integration tests'
+      env: GIMME_GO_VERSION=1.12.x FSC_PATH="/tmp/fsc-repo" TAGS=@FEATURE_ROLLOUT && ~@INPUT_FILTER"
+      before_script:
+        - mkdir -p $FSC_PATH
+        - pushd $FSC_PATH && git init && git fetch --depth=1 https://$CI_USER_TOKEN@github.com/optimizely/fullstack-sdk-compatibility-suite ${FSC_BRANCH:-master} && git checkout FETCH_HEAD && popd        
+      install:
+        - eval "$(gimme)"
+      script:
+        - ./scripts/run-fsc-tests.sh -f $FSC_PATH/features -d $FSC_PATH/features/support/datafiles/ -t "$TAGS"
```
mikeng13: @yasirfolio3 please let me know which tags to add in TravisJob env variable. Is the key going to be just `TAGS`? I can add it tonight so you can continue your work

yasirfolio3: Hey @mikeng13 , i have added `TAGS` as env in travis settings, only 1 test is failing with the added tags which is due to activate listener.

scripts/run-fsc-tests.sh
```diff
@@ -0,0 +1,70 @@
...
+
+set -ex
+GO_FEATUREFILES_PATH="$(pwd)/tests/integration/features"
+rm -rf $GO_FEATUREFILES_PATH
+mkdir -p $GO_FEATUREFILES_PATH
+cp -r $FEATURE_FILES_PATH $GO_FEATUREFILES_PATH
+
+export DATAFILES_DIR="$DATAFILES_PATH"
+go test -v $(pwd)/tests/integration --godog.tags="$TAG_FILTER_EXPRESSION" 
+echo "Ready for testing."
```
mikeng13: Shouldn't this go before we run `go test`?

PR #154, opened 2019-10-29
ci(srcclr): adding source clear

msohailhussain:


msohailhussain:
@juancarlostong Can you please take a look at this PR, I want not to add separate stage only for source clear.

PR #155, opened 2019-10-30
fix(ExperimentWhitelistService): Fix finding variation by key

mjc1283:
Previously, `GetDecision` of `ExperimentWhitelistService` was incorrectly using variation key to look up variations in the `Variations` map of an `Experiment` struct, in which keys are variation IDs, not variation keys. 

With this fix, we iterate over variations to find the one with the desired key. To avoid needing to search all variations, we can add a map keyed by variation key as a follow up.

### DIFF
--- a/pkg/decision/experiment_whitelist_service.go
+++ b/pkg/decision/experiment_whitelist_service.go
@@ -49,8 +49,8 @@ func (s ExperimentWhitelistService) GetDecision(decisionContext ExperimentDecisi
 
 	// TODO(Matt): Add a VariationsByKey map to the Experiment struct, and use it to look up Variation by key
 	for _, variation := range decisionContext.Experiment.Variations {
+		variation := variation
 		if variation.Key == variationKey {
-			variation := variation
 			decision.Reason = reasons.WhitelistVariationAssignmentFound
 			decision.Variation = &variation
 			return decision, nil
--- a/pkg/decision/experiment_whitelist_service.go
+++ b/pkg/decision/experiment_whitelist_service.go
@@ -47,13 +47,14 @@ func (s ExperimentWhitelistService) GetDecision(decisionContext ExperimentDecisi
 		return decision, nil
 	}
 
-	variation, ok := decisionContext.Experiment.Variations[variationKey]
-	if !ok {
-		decision.Reason = reasons.InvalidWhitelistVariationAssignment
-		return decision, nil
+	for _, variation := range decisionContext.Experiment.Variations {
+		if variation.Key == variationKey {
+			decision.Reason = reasons.WhitelistVariationAssignmentFound
+			decision.Variation = &variation
+			return decision, nil
+		}
 	}
 
-	decision.Reason = reasons.WhitelistVariationAssignmentFound
-	decision.Variation = &variation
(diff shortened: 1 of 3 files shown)


pkg/decision/experiment_whitelist_service.go
```diff
@@ -47,13 +47,16 @@ func (s ExperimentWhitelistService) GetDecision(decisionContext ExperimentDecisi
 		return decision, nil
 	}
 
-	variation, ok := decisionContext.Experiment.Variations[variationKey]
-	if !ok {
-		decision.Reason = reasons.InvalidWhitelistVariationAssignment
-		return decision, nil
+	// TODO(Matt): Add a VariationsByKey map to the Experiment struct, and use it to look up Variation by key
+	for _, variation := range decisionContext.Experiment.Variations {
+		variation := variation
```
mikeng13: Why is this reassignment necessary? Can't you just use the variation assigned by the loop iterator?

mjc1283: The linter forces it https://github.com/kyoh86/scopelint

mikeng13: Oh... @pawels-optimizely thoughts on this? I won't block the merge on this 🙂

pkg/decision/experiment_whitelist_service_test.go
```diff
@@ -122,6 +122,23 @@ func (s *ExperimentWhitelistServiceTestSuite) TestNoExperimentInDecisionContext(
 	s.Nil(decision.Variation)
 }
 
+// Test an experiment whose key differs from its id
+func (s *ExperimentWhitelistServiceTestSuite) TestWhitelistIncludesDecisionForExpWithKey() {
```
mikecdavis: Not sure we need a _new_ test for this or if we update the previous test fixture to have a differing id, key pair.

pkg/decision/experiment_whitelist_service.go
mikecdavis: Appears to be a valid catch for potentially hard to track down bugs. I also just learned something about go range loops :) In this particular case I believe we were safe since we're immediately returning with the current `variation`, but might as well be in the habit of "pinning".

pkg/decision/helpers_test.go
```diff
@@ -242,3 +242,21 @@ var testExpWhitelist = entities.Experiment{
 		"test_user_2": "2230",
 	},
 }
+
+// Experiment with a whitelist, and variation ids are not the same as keys
+const testExpWhitelistKey2 = "test_experiment_whitelist_2"
```
mikecdavis: Similar comment above about not needing a new fixture.

pkg/decision/experiment_whitelist_service_test.go
mjc1283: Good point, I'll delete the new test and change the old test data.

pkg/decision/experiment_whitelist_service.go
mikecdavis: @pawels-optimizely The assignment is not just needed for go routines, that is just a common place it can manifest itself. See the `copies` example in scopelint readme.

PR #156, opened 2019-10-30
feat(gherkin-track): Added new steps and support for testing track API.

yasirfolio3:
### Summary
- Added steps to process track API.
- General fixes.

### DIFF
--- a/tests/integration/main_test.go
+++ b/tests/integration/main_test.go
@@ -69,4 +69,5 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^the number of dispatched events is (\d+)$`, context.TheNumberOfDispatchedEventsIs)
 	s.Step(`^there are no dispatched events$`, context.ThereAreNoDispatchedEvents)
 	s.Step(`^dispatched events payloads include$`, context.DispatchedEventsPayloadsInclude)
+	s.Step(`^payloads of dispatched events don\'t include decisions$`, context.PayloadsOfDispatchedEventsDontIncludeDecisions)
 }
--- a/tests/integration/optlyplugins/proxy_dispatcher.go
+++ b/tests/integration/optlyplugins/proxy_dispatcher.go
@@ -43,8 +43,3 @@ func (d *ProxyEventDispatcher) GetEvents() []event.Batch {
 	}
 	return d.events
 }
-
-// ClearEvents deletes dispatched events
-func (d *ProxyEventDispatcher) ClearEvents() {
-	d.events = nil
-}
--- a/tests/integration/optlyplugins/proxy_dispatcher.go
+++ b/tests/integration/optlyplugins/proxy_dispatcher.go
@@ -43,3 +43,8 @@ func (d *ProxyEventDispatcher) GetEvents() []event.Batch {
 	}
 	return d.events
 }
+
+// ClearEvents deletes dispatched events
+func (d *ProxyEventDispatcher) ClearEvents() {
+	d.events = nil
+}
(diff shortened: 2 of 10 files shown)


tests/integration/models/get_enabled_features_params.go
```diff
@@ -16,8 +16,8 @@
 
 package models
 
-// GetEnabledFeaturesParams represents params required for GetEnabledFeatures API
-type GetEnabledFeaturesParams struct {
+// GetEnabledFeaturesRequestParams represents params required for GetEnabledFeatures API
+type GetEnabledFeaturesRequestParams struct {
```
msohailhussain: why you changed names?

tests/integration/main_test.go
```diff
@@ -69,4 +69,5 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^the number of dispatched events is (\d+)$`, context.TheNumberOfDispatchedEventsIs)
 	s.Step(`^there are no dispatched events$`, context.ThereAreNoDispatchedEvents)
 	s.Step(`^dispatched events payloads include$`, context.DispatchedEventsPayloadsInclude)
+	s.Step(`^payloads of dispatched events don\'t include decisions$`, context.PayloadsOfDispatchedEventsDontIncludeDecisions)
```
msohailhussain: this is interesting, i just realized this step in steps.js

tests/integration/models/get_enabled_features_params.go
yasirfolio3: to follow the same naming scheme for all request params.

tests/integration/support/steps.go
```diff
@@ -56,6 +56,13 @@ func (c *ScenarioCtx) ListenerIsAdded(numberOfListeners int, ListenerName string
 func (c *ScenarioCtx) IsCalledWithArguments(apiName string, arguments *gherkin.DocString) error {
 	c.apiOptions.APIName = apiName
 	c.apiOptions.Arguments = arguments.Content
+
+	// Clearing old state of response, eventdispatcher and decision service
+	c.apiResponse = models.APIResponse{}
+	// Only required for unsessioned tests
+	c.clientWrapper.DecisionService.(*optlyplugins.TestCompositeService).ClearListenersCalled()
```
mikeng13: HOw come we need to clean up between calls? Can we just generate a new client wrapper or "get" a new client wrapper based on the requestID?

yasirfolio3: This is for scenarios where multiple api's are called one after another. In such cases, listeners_called, dispatched_events and response of former request needs to be cleared before the latter request is executed. We only recreate client-wrapper when step "the datafile is" is called so for the given case, same client-wrapper is used.

mikeng13: What I was saying is that we can use the `request-id` from the request params to cached/fetch instances of client wrappers and create the clientWrapper here when we call `IsCalledWithArguments`. That way we don't need to explicitly clean up these components. We already do this in other SDKs for multiple API calls

yasirfolio3: done, can you please review it now.

tests/integration/optlyplugins/proxy_dispatcher.go
```diff
@@ -43,3 +43,8 @@ func (d *ProxyEventDispatcher) GetEvents() []event.Batch {
 	}
 	return d.events
 }
+
+// ClearEvents deletes dispatched events
+func (d *ProxyEventDispatcher) ClearEvents() {
```
mikeng13: This is not used anymore

tests/integration/optlyplugins/test_composite_service.go
```diff
@@ -52,6 +52,11 @@ func (c *TestCompositeService) GetListenersCalled() []models.DecisionListener {
 	return c.listenersCalled
 }
 
+// ClearListenersCalled - Deletes listeners called
+func (c *TestCompositeService) ClearListenersCalled() {
```
mikeng13: This is no longer needed

yasirfolio3: both changes made.

PR #157, opened 2019-10-30
fix(mutual-exclusion-groups): Added missing groupMap implementation and GroupId to Experiment.

yasirfolio3:
### Summary
- Created groupMap in config.
- Added GroupId to Experiments inside groups.

### DIFF
--- a/pkg/config/datafileprojectconfig/mappers/experiment_test.go
+++ b/pkg/config/datafileprojectconfig/mappers/experiment_test.go
@@ -19,7 +19,7 @@ package mappers
 import (
 	"testing"
 
-	"github.com/json-iterator/go"
+	jsoniter "github.com/json-iterator/go"
 	datafileEntities "github.com/optimizely/go-sdk/pkg/config/datafileprojectconfig/entities"
 	"github.com/optimizely/go-sdk/pkg/entities"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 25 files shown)


pkg/config/datafileprojectconfig/mappers/group.go
```diff
@@ -0,0 +1,43 @@
...
+	for _, group := range rawGroups {
+		groupEntity := entities.Group{
+			ID:     group.ID,
+			Policy: group.Policy,
+		}
+		for _, allocation := range group.TrafficAllocation {
+			groupEntity.TrafficAllocation = append(groupEntity.TrafficAllocation, entities.Range(allocation))
+		}
+		groupMap[group.ID] = groupEntity
+		for _, experiment := range group.Experiments {
```
mikeng13: I think this is redundant since we already pass the merged experiments to the `MapExperiments` which takes care of assigning the group ID right?

yasirfolio3: experimentGroupMap is also passed to `MapExperiments` which uses the map to get groupId for experimentId.

yasirfolio3: https://github.com/optimizely/go-sdk/pull/157/files#diff-b405b44ec009d9817d391cea754df725R33

pkg/config/datafileprojectconfig/mappers/experiment_test.go
```diff
@@ -114,6 +117,53 @@ func TestMapExperiments(t *testing.T) {
...
+			"id": "11112"
+		  }
+	  ],
+	  "id":"11112"
+	}`
+
+	var rawExperiment datafileEntities.Experiment
+	var rawGroup datafileEntities.Group
+	var json = jsoniter.ConfigCompatibleWithStandardLibrary
+	json.Unmarshal([]byte(testExperimentString), &rawExperiment)
```
mikeng13: Sorry I didn't actually catch this earlier, but we shouldn't need to unmarshal the json string into the raw entities. We should just directly create a `datafileEntities.Experiment{}` with the properties that we care about

pkg/config/datafileprojectconfig/mappers/group_test.go
```diff
@@ -0,0 +1,61 @@
...
+			"entityId": "13",
+			"endOfRange": 4000
+		  }
+		],
+		"id": "14"
+	  }`
+
+	var rawGroup datafileEntities.Group
+	var json = jsoniter.ConfigCompatibleWithStandardLibrary
+	json.Unmarshal([]byte(testGroupString), &rawGroup)
```
mikeng13: Same comment about unmarshalling

pkg/config/datafileprojectconfig/mappers/experiment_test.go
yasirfolio3: Thanks, done.

PR #159, opened 2019-10-30
RFC: Experiment overrides

mjc1283:
This WIP PR introduces "experiment overrides" as a way of satisfying the same requirements that are covered by "forced variation" methods in older SDKs.

The approach is to allow the SDK consumer to maintain their own map, which specifies the desired set of forced variations. They can add or remove entries from this map at will. The SDK provides an option function to be used with `NewCompositeService` that accepts the overrides map, and can be used to create a decision service which will honor the overrides in the map any time an experiment decision is made.

Comments requested on the following:

- There are no affordances for concurrency. Should there be? Do we say that the client is not thread safe, and push all responsibility to the user?
- As option functions proliferate, we might introduce implicit ordering or state requirements that ruin their composability, and I see no testing in place to prevent this. Should we have a standard framework or test strategy to prevent problems?
- When using option functions to overwrite certain struct fields, we potentially waste work creating values for fields, only for them to be overwritten by the option func. This could turn into a performance problem. Is the best general approach to create a zero-valued struct first, then apply options, and then finally fill in remaining fields? How do we tell in general what fields remain uninitialized after applying options?

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/optimizely/go-sdk/pkg/client"
+	"github.com/optimizely/go-sdk/pkg/decision"
 	"github.com/optimizely/go-sdk/pkg/entities"
 	"github.com/optimizely/go-sdk/pkg/event"
 	"github.com/optimizely/go-sdk/pkg/logging"
@@ -67,4 +68,24 @@ func main() {
 		client.WithBatchEventProcessor(event.DefaultBatchSize, event.DefaultEventQueueSize, event.DefaultEventFlushInterval),
 	)
 	optimizelyClient.Close()
+
+	/************* Setting experiment overrides (a.k.a. "forced variations") ********************/
+	overrideKey := decision.OverrideKey{
+		Experiment: "aaaa",
+		User:       "Matt",
+	}
+	overrides := map[decision.OverrideKey]string{
+		overrideKey: "variation_1",
+	}
+	compositeService := decision.NewCompositeService(
+		sdkKey,
+		decision.WithExperimentOverrides(overrides),
+	)
+	optimizelyClient, _ = optimizelyFactory.Client(
+		client.WithDecisionService(compositeService),
+	)
+	// Optimizely client now has "variation_1" forced for user "Matt" in experiment "aaaa"
+	// The forced variation will work regardless of whether "aaaa" is an A/B test or a Feature Test.
+	fmt.Printf("Is feature enabled? %v\n", enabled)
+
 }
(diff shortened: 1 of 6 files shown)


pkg/decision/experiment_override_service.go
```diff
@@ -0,0 +1,74 @@
...
+		return decision, errors.New("decisionContext Experiment is nil")
+	}
+
+	variationKey, ok := s.Overrides[OverrideKey{Experiment: decisionContext.Experiment.Key, User: userContext.ID}]
+	if !ok {
+		decision.Reason = reasons.NoOverrideVariationForUser
+		return decision, nil
+	}
+
+	for _, variation := range decisionContext.Experiment.Variations {
```
mjc1283: Just like in whitelist service, it can be a TODO to introduce and use a by-variation-key map. This is just for demonstration.

examples/main.go
```diff
@@ -67,4 +68,24 @@ func main() {
...
+	}
+	overrides := map[decision.OverrideKey]string{
+		overrideKey: "variation_1",
+	}
+	compositeService := decision.NewCompositeService(
+		sdkKey,
+		decision.WithExperimentOverrides(overrides),
+	)
+	optimizelyClient, _ = optimizelyFactory.Client(
+		client.WithDecisionService(compositeService),
```
mikeng13: Since overrides are a first-party feature, I think it would be good to have an option func directly on the client factory to instantiate a composite service with the overrides instead of having to rely on instantiating a compositeService outside of the client.

pkg/decision/composite_service.go
```diff
@@ -35,16 +35,34 @@ type CompositeService struct {
 	notificationCenter         notification.Center
 }
 
+// CSOptionFunc allows customization of the CompositeService returned from NewCompositeService
+type CSOptionFunc func(*CompositeService)
+
+// WithExperimentOverrides applies the argument experiment overrides to a composite service
+func WithExperimentOverrides(experimentOverrides map[OverrideKey]string) CSOptionFunc {
```
mikeng13: Outside the scope of this PR, but I think it's a little cumbersome to have to add the option func in these two places. I've been thinking about simplifying the decision package a little by having this `composite_service` encompass the `[]ExperimentService` and `[]FeatureService` from the `composite_experiment_service` and `composite_feature_service` structs and get rid of those two classes entirely since those are just wrappers for the arrays and looping logic. What do you think?

pawels-optimizely:
@mjc1283: "Is the best general approach to create a zero-valued struct first, then apply options, and then finally fill in remaining fields?" - yes , that should be the main purpose of using option functions. Initializing by option functions should be implemented in that way so that ordering does not matter. Option functions should be mutually exclusive with initializing the struct fields. .

PR #160, opened 2019-10-30
feat(ups): User profile service interface and integration.

mikeng13:
# Summary
- Introduces `UserProfileService` interface
- Integrates user profile service into `ExperimentBucketerService`
- Missing factory param for user to pass in [WIP]

### DIFF
--- a/pkg/decision/entities.go
+++ b/pkg/decision/entities.go
@@ -63,3 +63,9 @@ type ExperimentDecision struct {
 	Decision
 	Variation *entities.Variation
 }
+
+// UserProfile contains past bucketing decisions made about a user
+type UserProfile struct {
+	ID                  string
+	ExperimentBucketMap map[string]map[string]string
+}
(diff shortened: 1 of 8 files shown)


pkg/decision/entities.go
```diff
@@ -63,3 +63,9 @@ type ExperimentDecision struct {
 	Decision
 	Variation *entities.Variation
 }
+
+// UserProfile contains past bucketing decisions made about a user
+type UserProfile struct {
+	ID                  string
+	ExperimentBucketMap map[string]map[string]string
```
mjc1283: It's better to use a struct as a multi-dimensional key, otherwise you have to worry about initializing all the nested maps. Check out the section "Key Types" here https://blog.golang.org/go-maps-in-action

mikeng13: Interesting point.. I think this makes sense if `variation_id` is all we cared about mapping to. It could very well be that we want other kinds of info captured in the nested map of `experiment_id --> { }`. Also, the key within the nested map is not variable, it will always be `variation_id` so I don't think that using a struct is appropriate in this case anyway.

mjc1283: It seems better to me to represent a set of fields known ahead of time using a struct, not a map.

mikeng13:
CLosing this in favor of a different approach.

PR #161, opened 2019-10-31
fix: support audience ids

mikeng13:
# Summary
- Fixes bug where `experiment.audienceIds` is not parsed into audience condition tree in the absence of `experiment.audienceConditions`
- Fixes bug where user attribute lookup returns error when attribute not found during condition matching, and thus results in the `not` condition being short-circuited to false

### DIFF
--- a/pkg/config/datafileprojectconfig/entities/entities.go
+++ b/pkg/config/datafileprojectconfig/entities/entities.go
@@ -42,7 +42,7 @@ type Experiment struct {
 	TrafficAllocation  []trafficAllocation `json:"trafficAllocation"`
 	AudienceIds        []string            `json:"audienceIds"`
 	ForcedVariations   map[string]string   `json:"forcedVariations"`
-	AudienceConditions interface{}         `json:"audienceConditions"`
+	AudienceConditions []interface{}       `json:"audienceConditions"`
 }
 
 // FeatureFlag represents a FeatureFlag object from the Optimizely datafile
--- a/pkg/config/datafileprojectconfig/mappers/experiment.go
+++ b/pkg/config/datafileprojectconfig/mappers/experiment.go
@@ -60,7 +60,7 @@ func mapExperiment(rawExperiment datafileEntities.Experiment) entities.Experimen
 	var err error
 	if rawExperiment.AudienceConditions == nil && len(rawExperiment.AudienceIds) > 0 {
 		audienceConditionTree, err = buildAudienceConditionTree(rawExperiment.AudienceIds)
-	} else {
+	} else if len(rawExperiment.AudienceConditions) > 0 {
 		audienceConditionTree, err = buildAudienceConditionTree(rawExperiment.AudienceConditions)
 	}
 	if err != nil {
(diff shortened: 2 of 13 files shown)


pkg/decision/evaluator/matchers/exact.go
```diff
@@ -34,23 +34,23 @@ func (m ExactMatcher) Match(user entities.UserContext) (bool, error) {
 	if stringValue, ok := m.Condition.Value.(string); ok {
 		attributeValue, err := user.GetStringAttribute(m.Condition.Name)
 		if err != nil {
-			return false, err
+			return false, nil
```
mikecdavis: Should we log that we're swallowing these errors?

pkg/config/datafileprojectconfig/mappers/experiment.go
```diff
@@ -56,7 +56,13 @@ func mapVariation(rawVariation datafileEntities.Variation) entities.Variation {
 
 // Maps the raw experiment entity from the datafile into an SDK Experiment entity
 func mapExperiment(rawExperiment datafileEntities.Experiment) entities.Experiment {
-	audienceConditionTree, err := buildAudienceConditionTree(rawExperiment.AudienceConditions)
+	var audienceConditionTree *entities.TreeNode
+	var err error
+	if rawExperiment.AudienceConditions == nil && len(rawExperiment.AudienceIds) > 0 {
+		audienceConditionTree, err = buildAudienceConditionTree(rawExperiment.AudienceIds)
```
mikecdavis: Do we have coverage on this path?

pkg/decision/evaluator/matchers/exact.go
mikeng13: the error here means that the user does not have that attribute key we are looking for. I can log it as debug but that might be a bit too noisy, wdyt?

pkg/config/datafileprojectconfig/mappers/experiment.go
mikeng13: Added it.

mikeng13:
> LGTM, but I'm not quite sure I understand why we needed to change from returning the `err`, to always returning `nil`. Minor comment on lack of coverage that could result in a future regression albeit that it should be covered by the compat suite. Perhaps add a follow-up ticket for that fixture?

I tried to explain this in the summary 😄. It boils down to the error being used to return `false` but that breaks the `not` operator since the attribute not being present means that `not [condition]` should evaluate to true.

pkg/decision/evaluator/matchers/exact.go
mikeng13: Created JIRA ticket to track this.

pkg/config/datafileprojectconfig/entities/entities.go
```diff
@@ -42,7 +42,7 @@ type Experiment struct {
 	TrafficAllocation  []trafficAllocation `json:"trafficAllocation"`
 	AudienceIds        []string            `json:"audienceIds"`
 	ForcedVariations   map[string]string   `json:"forcedVariations"`
-	AudienceConditions interface{}         `json:"audienceConditions"`
```
yasirfolio3: @mikeng13 this change is causing issues for datafiles with audience conditions defined as:
`"audienceConditions": "20413101795"`

PR #163, opened 2019-10-31
feat(decision): User profile service

mikeng13:
# Summary
- Introduces UserProfileService interface
- Introduces a PersistingExperimentService class that wraps over the ExperimentBucketerService and performs UPS lookup and save
- Integration into the `CompositeExperimentService` will be done in a follow-up PR

### DIFF
--- a/pkg/decision/composite_experiment_service.go
+++ b/pkg/decision/composite_experiment_service.go
@@ -34,16 +34,14 @@ type CompositeExperimentService struct {
 
 // NewCompositeExperimentService creates a new instance of the CompositeExperimentService
 func NewCompositeExperimentService() *CompositeExperimentService {
-	persistingExperimentService := NewPersistingExperimentService(NewExperimentBucketerService())
-
 	// These decision services are applied in order:
 	// 1. Whitelist
-	// 2. Bucketing (wrapped with user profile service)
+	// 2. Bucketing
 	// @TODO(mng): Prepend forced variation
 	return &CompositeExperimentService{
 		experimentServices: []ExperimentService{
 			NewExperimentWhitelistService(),
-			persistingExperimentService,
+			NewExperimentBucketerService(),
 		},
 	}
 }
--- a/pkg/decision/composite_experiment_service.go
+++ b/pkg/decision/composite_experiment_service.go
@@ -34,14 +34,16 @@ type CompositeExperimentService struct {
 
 // NewCompositeExperimentService creates a new instance of the CompositeExperimentService
 func NewCompositeExperimentService() *CompositeExperimentService {
+	persistingExperimentService := NewPersistingExperimentService(NewExperimentBucketerService())
+
 	// These decision services are applied in order:
 	// 1. Whitelist
-	// 2. Bucketing
+	// 2. Bucketing (wrapped with user profile service)
 	// @TODO(mng): Prepend forced variation
 	return &CompositeExperimentService{
 		experimentServices: []ExperimentService{
 			NewExperimentWhitelistService(),
-			NewExperimentBucketerService(),
+			persistingExperimentService,
 		},
 	}
 }
--- a/pkg/decision/entities.go
+++ b/pkg/decision/entities.go
@@ -67,6 +67,15 @@ type ExperimentDecision struct {
 // UserDecisionKey is used to access the saved decisions in a user profile
 type UserDecisionKey struct {
 	ExperimentID string
+	Field        string
+}
+
+// NewUserDecisionKey returns a new UserDecisionKey with the given experiment ID
+func NewUserDecisionKey(experimentID string) UserDecisionKey {
+	return UserDecisionKey{
+		ExperimentID: experimentID,
+		Field:        "variation_id",
+	}
 }
 
 // UserProfile represents a saved user profile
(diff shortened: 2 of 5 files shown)


pkg/decision/persisting_experiment_service.go
```diff
@@ -0,0 +1,119 @@
...
+
+// PESOptionFunc is used to extend the PersistingExperimentService with additional options
+type PESOptionFunc func(*PersistingExperimentService)
+
+// WithUserProfileService sets the user profile service on the persistingExperimentService instance
+func WithUserProfileService(userProfileService UserProfileService) PESOptionFunc {
+	return func(f *PersistingExperimentService) {
+		f.userProfileService = userProfileService
+	}
+}
```
mjc1283: Why is this optional? Seems like it should be required.

pkg/decision/composite_experiment_service.go
```diff
@@ -34,14 +34,16 @@ type CompositeExperimentService struct {
 
 // NewCompositeExperimentService creates a new instance of the CompositeExperimentService
 func NewCompositeExperimentService() *CompositeExperimentService {
+	persistingExperimentService := NewPersistingExperimentService(NewExperimentBucketerService())
```
mjc1283: Should we only apply the persisting wrapper when they use an option func and actually provide a user profile service? Otherwise, most of the time, this persisting experiment service is running, but not doing anything interesting.

mikeng13: Yeah you are right. Let me fix that.

pkg/decision/persisting_experiment_service.go
```diff
@@ -0,0 +1,119 @@
...
+		opt(persistingExperimentService)
+	}
+
+	return persistingExperimentService
+}
+
+// GetDecision returns the decision with the variation the user is bucketed into
+func (p PersistingExperimentService) GetDecision(decisionContext ExperimentDecisionContext, userContext entities.UserContext) (ExperimentDecision, error) {
+	var experimentDecision ExperimentDecision
+	var err error
```
pawels-optimizely: var experimentDecision ExperimentDecision
	var err error  -- these could be defined on the return level

pkg/decision/persisting_experiment_service.go
```diff
@@ -0,0 +1,119 @@
...
+	var userProfile UserProfile
+	// check to see if there is a saved decision for the user
+	if p.userProfileService != nil {
```
pawels-optimizely: seems like the whole logic depends on userProfileService being not null. Should we check for this at the beginning
and return right away ?

also it's a bit confusing to me that getSavedDecision returns profile, but I know why it's doing that.

pawels-optimizely:
```if p.userProfileService  == nil  {
	return p.experimentBucketedService.GetDecision(decisionContext, userContext)

}



experimentDecision, userProfile = p.getSavedDecision(decisionContext, userContext)
if experimentDecision.Variation != nil {
	return experimentDecision, nil
}

experimentDecision, err = p.experimentBucketedService.GetDecision(decisionContext, userContext)

p.saveDecision(userProfile, decisionContext.Experiment, experimentDecision)

return experimentDecision, err```

File: pkg/decision/entities.go [hunk 5]
```diff
@@ -63,3 +63,14 @@ type ExperimentDecision struct {
 	Decision
 	Variation *entities.Variation
 }
+
+// UserDecisionKey is used to access the saved decisions in a user profile
+type UserDecisionKey struct {
+	ExperimentID string
+}
```
mjc1283: Why use a struct here, as opposed to just a `string`?

pkg/decision/interface.go
```diff
@@ -39,3 +39,9 @@ type ExperimentService interface {
 type FeatureService interface {
 	GetDecision(decisionContext FeatureDecisionContext, userContext entities.UserContext) (FeatureDecision, error)
 }
+
+// UserProfileService is used to save and retrieve past bucketing decisions for users
+type UserProfileService interface {
+	Lookup(string) (UserProfile, error)
```
mjc1283: Do we need `Lookup` to return an error? The SDK internals don't have to care whether there was an error - that seems more like a concern of the implementation. Whoever implements this could track and handle errors on their end. Do they really need to route an error through the SDK logger like that?

From the SDK's perspective, it just needs a user profile for a given user id. We could require that a `UserProfile` is always returned, and say that when there is no saved decision for that user id (for whatever reason, error or not), then the implementation should return a `UserProfile` containing an empty `ExperimentBucketMap`.

pkg/decision/interface.go
```diff
@@ -39,3 +39,9 @@ type ExperimentService interface {
...
+	Save(UserProfile) error
```
mjc1283: Same comment above applies to returning an `error` here. Why does the SDK care?

pkg/decision/entities.go
mikeng13: Good catch. In our current user profile service implementation we use a combination of experimentID and the key `variation_id`. I don't remember why we decided to make that value a map instead of pointing from experimentID to variationID and I added the struct so we could maintain that relationship and maintain consistency with the other SDKs.

pkg/decision/interface.go
mikeng13: This is a good point. I'll remove the errors since we don't actually care about them.

PR #164, opened 2019-11-01
feat: Experiment override service

mjc1283:
## Summary

Add `ExperimentOverrideService`, a decision service that can pull variations from an `ExperimentOverrideStore`, which is an interface that returns variations for experiment key/user ID pairs.

Included is an implementation of `ExperimentOverrideStore` based on a map.

## Test plan

New unit tests

## JIRA Issues
https://optimizely.atlassian.net/browse/OASIS-5419

### DIFF
--- a/pkg/decision/experiment_override_service.go
+++ b/pkg/decision/experiment_override_service.go
@@ -30,7 +30,7 @@ var eosLogger = logging.GetLogger("ExperimentOverrideService")
 
 // ExperimentOverrideKey represents the user ID and experiment associated with an override variation
 type ExperimentOverrideKey struct {
-	Experiment, UserID string
+	ExperimentKey, UserID string
 }
 
 // OverrideStore provides read access to overrides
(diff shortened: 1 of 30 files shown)


pkg/decision/experiment_override_service.go
```diff
@@ -0,0 +1,93 @@
...
+		return decision, nil
+	}
+
+	// TODO(Matt): Add a VariationsByKey map to the Experiment struct, and use it to look up Variation by key
+	for _, variation := range decisionContext.Experiment.Variations {
+		variation := variation
+		if variation.Key == variationKey {
+			decision.Variation = &variation
+			decision.Reason = reasons.OverrideVariationAssignmentFound
+			eosLogger.Info(fmt.Sprintf("Override variation %v found for user %v", variationKey, userContext.ID))
```
mikeng13: I think this should be a debug level message

pkg/decision/experiment_override_service.go
```diff
@@ -0,0 +1,93 @@
...
+		return decision, errors.New("decisionContext Experiment is nil")
+	}
+
+	variationKey, ok := s.Overrides.GetVariation(ExperimentOverrideKey{ExperimentKey: decisionContext.Experiment.Key, UserID: userContext.ID})
+	if !ok {
+		decision.Reason = reasons.NoOverrideVariationAssignment
+		return decision, nil
+	}
+
+	// TODO(Matt): Add a VariationsByKey map to the Experiment struct, and use it to look up Variation by key
```
mikeng13: Probably a keyToID map and then you can just look up by ID

pawels-optimizely: is this TODO still needed ?

mjc1283: I think it's needed. Ideally we would not have to iterate over variations to find the one with a given key.

PR #165, opened 2019-11-04
fix(audience-conditions): fixed parsing for audience conditions.

yasirfolio3:
## Summary

- Fixed audienceConditions parsing when a leaf is provided instead of an array in the datafile.


yasirfolio3:
> Let's add a unit test to capture the regression

Unit test added, thanks.

PR #167, opened 2019-11-06
(fix): cleanup event processor

thomaszurkan-optimizely:
**Summary**
- When event batch size has been reached only start one batch event processing go routine.
- When queue size is met, log a message and do not add to the queue.
- Duration used was setting the time too far into the future by multiplying by second and then by milliseconds.  Thank Mike D for catching this one.

**Testing**
This PR has unit tests for all the fixes above.


pkg/event/processor.go
```diff
@@ -142,13 +150,26 @@ func (p *BatchEventProcessor) Start(exeCtx utils.ExecutionCtx) {
...
-		go func() {
-			p.FlushEvents()
-		}()
+		pLogger.Warning("MaxQueueSize has been met. Discarding event")
+		return
 	}
+
+	p.Q.Add(event)
+
+	if p.Q.Size() >= p.BatchSize {
```
mikecdavis: Can we flatten this out:
```go
if p.Q.Size() < p.BatchSize || p.processing {
    return
}

p.processing = true
go func() {
    p.FlushEvents()
    p.processing = false
}()
```

pkg/event/processor.go
```diff
@@ -44,6 +44,7 @@ type BatchEventProcessor struct {
 	Mux             sync.Mutex
 	Ticker          *time.Ticker
 	EventDispatcher Dispatcher
+	processing      bool
```
mikecdavis: This is like a poor man's Mutex :)

thomaszurkan-optimizely: you noticed that.  Yeah, I set it and test it in the same thread and then set it back in the other thread.  So, there is really no chance of collision.  However, go lint disagreed.  So, I went ahead and created a new util struct called AtomicProperty.

pkg/event/processor.go
```diff
@@ -142,13 +148,26 @@ func (p *BatchEventProcessor) Start(exeCtx utils.ExecutionCtx) {
...
-			p.FlushEvents()
-		}()
+		pLogger.Warning("MaxQueueSize has been met. Discarding event")
+		return
 	}
+
+	p.Q.Add(event)
+
+	if p.Q.Size() >= p.BatchSize {
+		if processing, ok := p.processing.Get().(bool); ok && !processing {
```
mikecdavis: Perhaps you can use a [semaphore](https://godoc.org/golang.org/x/sync/semaphore).
```go
if p.sem.TryAquire(1) {
    go func() {
        p.FlushEvents()
        p.sem.Release(1)
    }()
}
```

ref: https://medium.com/@deckarep/gos-extended-concurrency-semaphores-part-1-5eeabfa351ce

pkg/event/processor.go
```diff
@@ -123,6 +128,11 @@ func NewBatchEventProcessor(options ...BPOptionConfig) *BatchEventProcessor {
 		p.BatchSize = DefaultBatchSize
 	}
 
+	if p.BatchSize > p.MaxQueueSize {
+		pLogger.Warning("Batch size is larger than queue size.  Swapping")
+		p.BatchSize, p.MaxQueueSize = p.MaxQueueSize, p.BatchSize
```
mikecdavis: This is a bit sneaky and I'm not sure we want to swap parameters like this.

thomaszurkan-optimizely: actually, it was a lint suggestion.  i love it!  :)

PR #168, opened 2019-11-06
feat(client): Allow user profile service and overrides to be passed in.

mikeng13:
# Summary
- Introduces options for User profile service and Experiment overrides store

# Tests
- Unit (pending)

### DIFF
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -46,13 +46,13 @@ type OptionFunc func(*OptimizelyFactory)
 
 // Client gets client and sets some parameters
 func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClient, error) {
-	if f.SDKKey == "" && f.Datafile == nil && f.configManager == nil {
-		return nil, errors.New("unable to instantiate client: no project config manager, SDK key, or a Datafile provided")
-	}
-
 	// extract options
 	for _, opt := range clientOptions {
 		opt(&f)
+	}
+
+	if f.SDKKey == "" && f.Datafile == nil && f.configManager == nil {
+		return nil, errors.New("unable to instantiate client: no project config manager, SDK key, or a Datafile provided")
 	}
 
 	var executionCtx utils.ExecutionCtx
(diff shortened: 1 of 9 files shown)


pkg/decision/composite_experiment_service.go
```diff
@@ -27,23 +27,57 @@ import (
...
+	return func(f *CompositeExperimentService) {
+		f.overrideStore = overrideStore
+	}
+}
+
 // CompositeExperimentService bridges together the various experiment decision services that ship by default with the SDK
 type CompositeExperimentService struct {
 	experimentServices []ExperimentService
+	overrideStore      ExperimentOverrideStore
+	userProfileService UserProfileService
```
mjc1283: Nit: it's a little weird that these fields are just use to move values from the option functions to a place where `NewCompositeExperimentService` can use them. Not sure there is a better alternative though.

pkg/decision/composite_service.go
```diff
@@ -35,16 +35,35 @@ type CompositeService struct {
...
+		opts(compositeService)
 	}
+
+	if compositeService.compositeExperimentService == nil {
+		compositeService.compositeExperimentService = NewCompositeExperimentService()
+	}
+
+	if compositeService.compositeFeatureService == nil {
+		compositeService.compositeFeatureService = NewCompositeFeatureService(compositeService.compositeExperimentService)
+	}
```
mjc1283: Should we just assign this instead of check for `nil`, since there is not yet an option func that allows passing in `compositeFeatureService`?

mjc1283:
@pawels-optimizely It seems more complex/potentially confusing to allow "some, but not all, fields are initialized prior to calling option functions", as opposed to initializing all fields before options, or none.

pkg/client/factory.go
```diff
@@ -32,35 +32,72 @@ import (
...
+	userProfileService decision.UserProfileService
+	overrideStore      decision.ExperimentOverrideStore
 }
 
 // OptionFunc is a type to a proper func
-type OptionFunc func(*OptimizelyClient)
+type OptionFunc func(*OptimizelyFactory)
 
 // Client gets client and sets some parameters
 func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClient, error) {
```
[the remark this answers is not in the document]
mikeng13: Yeah because this method does not need to make changes to the `OptimizelyFactory` itself so it's safer just to make it a value receiver

pkg/decision/composite_experiment_service.go
mikeng13: Yeah it was weird to me but at least they are private

PR #171, opened 2019-11-07
fix: Update built-in map-based ExperimentOverrideStore implementation to be concurrently usable

mjc1283:
The built-in default implementation of `ExperimentOverrideStore` should be production ready, so it should be safe to use concurrently.

In this PR:
- The default implementation is renamed to `MapExperimentOverridesStore`
- A `sync.RWLock` is added to `MapExperimentOverridesStore`
- A `SetVariation` method is added for setting an override, which uses the mutex
- The `GetVariation` method is updated to use the mutex
- A new test is added that calls `GetVariation` and `SetVariation` concurrently from several goroutines

### DIFF
--- a/pkg/decision/experiment_override_service_test.go
+++ b/pkg/decision/experiment_override_service_test.go
@@ -18,6 +18,7 @@
 package decision
 
 import (
+	"sync"
 	"testing"
 
 	"github.com/optimizely/go-sdk/pkg/decision/reasons"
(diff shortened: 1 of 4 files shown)


mjc1283:
@thomaszurkan-optimizely Based on reading this: https://golang.org/pkg/sync/#Map I gather that `sync.Map`, gives you better performance for certain use cases and frees you from needing to implement locking, but at the cost of type safety. Implementing locking is easy, and I don't know that we will always match one of those use cases, so it didn't seem worth giving up type safety. Let me know if you agree.

Also, aside from all that, can we even use it if we want our SDK to be usable in version <1.9?

pkg/decision/experiment_override_service_test.go
```diff
@@ -125,13 +126,63 @@ func (s *ExperimentOverrideServiceTestSuite) TestInvalidVariationInOverride() {
...
+	user1Decision, _ := s.overrideService.GetDecision(testDecisionContext, entities.UserContext{
+		ID: "test_user_1",
+	})
+	user2Decision, _ := s.overrideService.GetDecision(testDecisionContext, entities.UserContext{
+		ID: "test_user_2",
+	})
+	user3Decision, _ := s.overrideService.GetDecision(testDecisionContext, entities.UserContext{
+		ID: "test_user_3",
+	})
+	s.NotNil(user1Decision.Variation)
```
mikeng13: Can you also assert that the decision variation key matches what you expected?

PR #175, opened 2019-11-11
feat(audience-evaluation): Added support for parsing map for leaf condition.

yasirfolio3:
### Summary
- Added support for parsing map for leaf condition.
- Made further methods from `buildConditionTree` method to avoid cyclo-complexity warnings from linter.

### DIFF
--- a/pkg/config/datafileprojectconfig/mappers/condition_trees.go
+++ b/pkg/config/datafileprojectconfig/mappers/condition_trees.go
@@ -31,15 +31,9 @@ var json = jsoniter.ConfigCompatibleWithStandardLibrary
 // Takes the conditions array from the audience in the datafile and turns it into a condition tree
 func buildConditionTree(conditions interface{}) (conditionTree *entities.TreeNode, retErr error) {
 
-	var parsedConditions interface{}
-	switch v := conditions.(type) {
-	case string:
-		if err := json.Unmarshal([]byte(v), &parsedConditions); err != nil {
-			retErr = err
-			return
-		}
-	default:
-		parsedConditions = conditions
+	parsedConditions, retErr := validateConditions(conditions)
+	if retErr != nil {
+		return
 	}
 
 	value := reflect.ValueOf(parsedConditions)
@@ -74,17 +68,10 @@ func buildConditionTree(conditions interface{}) (conditionTree *entities.TreeNod
 					continue
 
 				case map[string]interface{}:
-					jsonBody, err := json.Marshal(typedV)
-					if err != nil {
+					if err := createLeafCondition(value, n); err != nil {
 						retErr = err
 						return
 					}
-					condition := entities.Condition{}
-					if err := json.Unmarshal(jsonBody, &condition); err != nil {
-						retErr = err
-						return
-					}
-					n.Item = condition
 				}
 
 				root.Nodes = append(root.Nodes, n)
@@ -93,19 +80,12 @@ func buildConditionTree(conditions interface{}) (conditionTree *entities.TreeNod
 			}
 		case reflect.Map:
 			typedV := v.Interface()
-			if _, ok := typedV.(map[string]interface{}); ok {
+			if v, ok := typedV.(map[string]interface{}); ok {
 				n := &entities.TreeNode{}
-				jsonBody, err := json.Marshal(typedV)
-				if err != nil {
+				if err := createLeafCondition(v, n); err != nil {
 					retErr = err
 					return
 				}
-				condition := entities.Condition{}
-				if err := json.Unmarshal(jsonBody, &condition); err != nil {
-					retErr = err
-					return
-				}
-				n.Item = condition
 				n.Operator = "or"
 				root.Operator = n.Operator
 				root.Nodes = append(root.Nodes, n)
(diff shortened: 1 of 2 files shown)


pkg/config/datafileprojectconfig/mappers/condition_trees_test.go
```diff
@@ -159,3 +159,28 @@ func TestBuildConditionTreeSimpleAudienceCondition(t *testing.T) {
 	}
 	assert.Equal(t, expectedConditionTree, conditionTree)
 }
+
+func TestBuildConditionTreeWithLeafNode(t *testing.T) {
+	conditionString := "{ \"type\": \"custom_attribute\", \"name\": \"s_foo\", \"match\": \"exact\", \"value\": \"foo\" }"
```
pawels-optimizely: @mikeng13 , I thought this case cannot be created by default, our app does not allow for that, is this the case ? 
if so then we should not support it.

mikeng13: Yeah looks like this case is supported. I wasn’t aware of it before but we have an integration test around it which all other SDKs support

mikeng13: Yeah looks like this case is supported. I wasn’t aware of it before but we have an integration test around it which all other SDKs support

pkg/config/datafileprojectconfig/mappers/condition_trees.go
```diff
@@ -93,22 +81,64 @@ func buildConditionTree(conditions interface{}) (conditionTree *entities.TreeNod
 		}
 	}
 
-	populateConditions(value, conditionTree)
+	// Check for leaf conditions
+	if value.Kind() == reflect.Map {
+		typedV := value.Interface()
+		if v, ok := typedV.(map[string]interface{}); ok {
+			n := &entities.TreeNode{}
+			if err := createLeafCondition(v, n); err != nil {
```
mikeng13: Is there a need to create the node out here and pass it in by reference? Why not just have the createLeafCondition instantiate and populate the node and return it?

pkg/config/datafileprojectconfig/mappers/condition_trees.go
```diff
@@ -93,22 +81,64 @@ func buildConditionTree(conditions interface{}) (conditionTree *entities.TreeNod
...
 	if conditionTree.Nodes == nil && conditionTree.Operator == "" {
 		retErr = errEmptyTree
 		conditionTree = nil
 	}
 	return conditionTree, retErr
 }
 
+// Validates conditions for audience in the datafile
+func validateConditions(conditions interface{}) (parsedConditions interface{}, retErr error) {
+	switch v := conditions.(type) {
```
mikeng13: This doesn’t really validate conditions. It’s just trying to parse it into an interface

pkg/config/datafileprojectconfig/mappers/condition_trees_test.go
```diff
@@ -159,3 +159,28 @@ func TestBuildConditionTreeSimpleAudienceCondition(t *testing.T) {
...
+	var conditions interface{}
+	json.Unmarshal([]byte(conditionString), &conditions)
+	conditionTree, err := buildConditionTree(conditions)
+	if err != nil {
```
mikeng13: Instead of doing this can you assert no error? Seems a little more straightforward

pkg/config/datafileprojectconfig/mappers/condition_trees.go
pawels-optimizely: yeah, maybe the name is not meaningful, but it just parses this into interface

pkg/config/datafileprojectconfig/mappers/condition_trees.go
pawels-optimizely: createLeafCondition is just a wrapper around existing logic. I don't have any preferences here. if you return a node, you still need to return an error to find out if you need to do ```node.Item = condition ``` assignment and normal flow.

pkg/config/datafileprojectconfig/mappers/condition_trees_test.go
pawels-optimizely: agree

pkg/config/datafileprojectconfig/mappers/condition_trees.go
yasirfolio3: it is mainly because `createLeafCondition` is being used on one other place aswell where reference passing is required.

pkg/config/datafileprojectconfig/mappers/condition_trees_test.go
yasirfolio3: done 👍

pkg/config/datafileprojectconfig/mappers/condition_trees.go
yasirfolio3: changed method name to `parseConditions`.

pkg/config/datafileprojectconfig/mappers/condition_trees.go
mikeng13: Ah in that makes sense

PR #178, opened 2019-11-12
feat: exposing variable on FeatureDecisionContext

pawels-optimizely:
related to https://optimizely.atlassian.net/browse/OASIS-5597

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -214,12 +214,12 @@ func (o *OptimizelyClient) GetFeatureVariableString(featureKey, variableKey stri
 // GetFeatureVariable returns feature as a string along with it's associated type
 func (o *OptimizelyClient) GetFeatureVariable(featureKey, variableKey string, userContext entities.UserContext) (value string, valueType entities.VariableType, err error) {
 
-	_, featureDecision, err := o.getFeatureDecision(featureKey, variableKey, userContext)
+	featureDecisionContext, featureDecision, err := o.getFeatureDecision(featureKey, variableKey, userContext)
 	if err != nil {
 		return "", "", err
 	}
 
-	variable := featureDecision.Variable
+	variable := featureDecisionContext.Variable
 
 	if featureDecision.Variation != nil {
 		if v, ok := featureDecision.Variation.Variables[variable.ID]; ok && featureDecision.Variation.FeatureEnabled {
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -96,7 +96,7 @@ func (o *OptimizelyClient) IsFeatureEnabled(featureKey string, userContext entit
 		}
 	}()
 
-	decisionContext, featureDecision, err := o.getFeatureDecision(featureKey, userContext)
+	decisionContext, featureDecision, err := o.getFeatureDecision(featureKey, "", userContext)
 	if err != nil {
 		logger.Error("received an error while computing feature decision", err)
 		return result, err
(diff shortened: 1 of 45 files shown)


pkg/client/client.go
```diff
@@ -214,15 +214,12 @@ func (o *OptimizelyClient) GetFeatureVariableString(featureKey, variableKey stri
 // GetFeatureVariable returns feature as a string along with it's associated type
 func (o *OptimizelyClient) GetFeatureVariable(featureKey, variableKey string, userContext entities.UserContext) (value string, valueType entities.VariableType, err error) {
 
-	decisionContext, featureDecision, err := o.getFeatureDecision(featureKey, userContext)
+	_, featureDecision, err := o.getFeatureDecision(featureKey, variableKey, userContext)
```
mikeng13: You should get the var from the context so you don't have to put it in the decision

pkg/decision/entities.go
```diff
@@ -56,6 +57,7 @@ type FeatureDecision struct {
 	Source     Source
 	Experiment entities.Experiment
 	Variation  *entities.Variation
+	Variable   *entities.Variable
```
mikeng13: Like mentioned above, I don't think you need to pass it in the decision. It's enought that it's in the context

pkg/decision/composite_service.go
```diff
@@ -69,13 +69,24 @@ func (s CompositeService) GetFeatureDecision(featureDecisionContext FeatureDecis
...
+			sourceInfo["variationKey"] = featureDecision.Variation.Key
+		}
+
 		featureInfo := map[string]interface{}{
-			"feature_key":     featureDecisionContext.Feature.Key,
-			"feature_enabled": false,
-			"source":          featureDecision.Source,
+			"featureKey":     featureDecisionContext.Feature.Key,
+			"featureEnabled": false,
+			"variableKey":    featureDecision.Variable.Key,
```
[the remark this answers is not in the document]
pawels-optimizely: this is exactly what java implementation has, do you want to change that in go sdk ?

pkg/client/client.go
pawels-optimizely: addressed

pkg/decision/entities.go
pawels-optimizely: addressed

pkg/decision/composite_service.go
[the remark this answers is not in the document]
pawels-optimizely: addressed

pkg/decision/composite_service_test.go
```diff
@@ -90,6 +97,37 @@ func (s *CompositeServiceFeatureTestSuite) TestDecisionListeners() {
 	s.Equal(numberOfCalls, 1)
 }
 
+func (s *CompositeServiceFeatureTestSuite) TestDecisionListenersNotificationContent() {
```
mikeng13: nit: `TestDecisionListenersNotificationWithVariable`

pkg/client/client_test.go
```diff
@@ -1480,7 +1503,7 @@ func TestGetAllFeatureVariablesWithoutFeature(t *testing.T) {
 }
 
 // Helper Methods
-func getTestFeatureDecision(experiment entities.Experiment, variation entities.Variation, decisionMade bool) decision.FeatureDecision {
+func getTestFeatureDecision(experiment entities.Experiment, variable entities.Variable, variation entities.Variation, decisionMade bool) decision.FeatureDecision {
```
mikeng13: This change is not needed anymore

pkg/decision/composite_service.go
```diff
@@ -69,13 +70,50 @@ func (s CompositeService) GetFeatureDecision(featureDecisionContext FeatureDecis
...
+			variableValue := variable.DefaultValue
+			if v, ok := featureDecision.Variation.Variables[variable.ID]; ok && featureDecision.Variation.FeatureEnabled {
+				variableValue = v.Value
+			}
+			var convertedValue interface{}
+			var e error
+
+			switch variable.Type {
+			case entities.String:
+				convertedValue, e = variableValue, nil
```
mikeng13: Does `e` need to be explicitly set to `nil`? Also, can we skip this case and instead always set `convertedValue` to the `variableValue` and then it gets overwritten if any of the cases below match? That way we always have a string value as backup/default in case we ever encounter an unknown type. Additionally we don't need to do the `if e != nil` check and conditionally assign since it will already point to the `variableValue`

pkg/decision/composite_service_test.go
mikeng13: Please also add a test case where we do not have variable info in the notification content

pkg/client/client_test.go
pawels-optimizely: addressed

pkg/decision/composite_service_test.go
pawels-optimizely: addressed

pkg/decision/composite_service.go
pawels-optimizely: error does not have to be initialized to nil explicitly, I just followed the pattern :) I will change it.  I will change it, but I don't think we can ever get unknown type, as we defined a custom type. We need to check for error, otherwise the converted value will be assigned to defaults (0.0, 0, false) in case of conversion problems

pkg/decision/composite_service_test.go
pawels-optimizely: addressed

PR #179, opened 2019-11-13
fix: don't return error when there is a good projectConfig in polling manager

pawels-optimizely:


pkg/config/polling_manager.go
```diff
@@ -94,7 +94,10 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
 
 		if e != nil {
 			cmLogger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
```
mikeng13: I think we can tone this down to a `Warning`. @mjc1283 or @mikecdavis what do you guys think?

mikeng13: Chatted with @mjc1283 offline and this is in line with what the other SDKs do.

PR #180, opened 2019-11-13
fix(decision): Logs produced by the various decision services.

mikeng13:
# Summary
- Fixes where and how we log various messages during the decision making process:
- 1. Gets rid of the `No decision made error` and instead returns the `decision` and `err` of the last decision maker
- 2. Logs debug messages in both feature experiment service as well as rollout service for clarity
- 3. Only logs (debug) bucketing ID if it is different from the user ID
- 4. Refactored the rollout_service_test to use suites

# Tests
- Unit

### DIFF
--- a/pkg/decision/composite_feature_service.go
+++ b/pkg/decision/composite_feature_service.go
@@ -51,7 +51,7 @@ func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionCont
 			cfLogger.Debug(fmt.Sprintf("%v", err))
 		}
 
-		if featureDecision.Variation != nil && featureDecision.Variation.FeatureEnabled && err == nil {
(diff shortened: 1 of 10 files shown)


pkg/decision/composite_feature_service.go
```diff
@@ -50,7 +50,8 @@ func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionCont
 		if err != nil {
 			cfLogger.Debug(fmt.Sprintf("%v", err))
 		}
-		if featureDecision.Variation != nil && err == nil {
+
+		if featureDecision.Variation != nil && featureDecision.Variation.FeatureEnabled && err == nil {
```
mjc1283: This seems wrong - why only return the decision if the feature is enabled?

PR #187, opened 2019-11-14
(chore): benchmark different event processor configurations

thomaszurkan-optimizely:
- default queue size should allow for a app to take a load out of the box. Changed to 1k
- benchmark different queue sizes, queue types, and batch sizes.
- discovered a potential race condition with logger.  that is fixed here is as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,17 +3,13 @@ module github.com/optimizely/go-sdk
 go 1.12
 
 require (
-	github.com/DATA-DOG/godog v0.7.13
-	github.com/google/go-cmp v0.3.1
 	github.com/google/uuid v1.1.1
 	github.com/json-iterator/go v1.1.7
-	github.com/optimizely/subset v0.0.0-20191114204000-ca68201b1be8
 	github.com/pkg/profile v1.3.0
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/testify v1.4.0
 	github.com/twmb/murmur3 v1.0.0
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
-	gopkg.in/yaml.v3 v3.0.0-20191120175047-4206685974f2
 )
 
 // Work around issue wtih git.apache.org/thrift.git
--- a/go.mod
+++ b/go.mod
@@ -3,13 +3,17 @@ module github.com/optimizely/go-sdk
 go 1.12
 
 require (
+	github.com/DATA-DOG/godog v0.7.13
+	github.com/google/go-cmp v0.3.1
 	github.com/google/uuid v1.1.1
 	github.com/json-iterator/go v1.1.7
+	github.com/optimizely/subset v0.0.0-20191114204000-ca68201b1be8
 	github.com/pkg/profile v1.3.0
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/testify v1.4.0
 	github.com/twmb/murmur3 v1.0.0
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
+	gopkg.in/yaml.v3 v3.0.0-20191120175047-4206685974f2
 )
 
 // Work around issue wtih git.apache.org/thrift.git
--- a/pkg/event/processor.go
+++ b/pkg/event/processor.go
@@ -155,17 +155,17 @@ func (p *BatchEventProcessor) Start(exeCtx utils.ExecutionCtx) {
 }
 
 // ProcessEvent processes the given impression event
-func (p *BatchEventProcessor) ProcessEvent(event UserEvent) {
+func (p *BatchEventProcessor) ProcessEvent(event UserEvent) bool {
 
 	if p.Q.Size() >= p.MaxQueueSize {
 		pLogger.Warning("MaxQueueSize has been met. Discarding event")
-		return
+		return false
 	}
 
 	p.Q.Add(event)
 
 	if p.Q.Size() < p.BatchSize {
-		return
+		return true
 	}
 
 	if p.processing.TryAcquire(1) {
@@ -178,6 +178,7 @@ func (p *BatchEventProcessor) ProcessEvent(event UserEvent) {
 		}()
 	}
 
+	return true
 }
 
 // EventsCount returns size of an event queue
--- a/pkg/event/processor.go
+++ b/pkg/event/processor.go
@@ -33,7 +33,7 @@ import (
 
 // Processor processes events
 type Processor interface {
-	ProcessEvent(event UserEvent)
+	ProcessEvent(event UserEvent) bool
 }
 
 // BatchEventProcessor is used out of the box by the SDK
--- a/pkg/event/processor.go
+++ b/pkg/event/processor.go
@@ -53,7 +53,7 @@ type BatchEventProcessor struct {
 const DefaultBatchSize = 10
 
 // DefaultEventQueueSize holds the default value for the event queue size
-const DefaultEventQueueSize = 100
+const DefaultEventQueueSize = 2000
 
 // DefaultEventFlushInterval holds the default value for the event flush interval
 const DefaultEventFlushInterval = 30 * time.Second
--- a/pkg/event/processor_test.go
+++ b/pkg/event/processor_test.go
@@ -485,14 +485,19 @@ func TestBenchmarkProcessor(t *testing.T) {
 
 	result := testing.Benchmark(benchmarkProcessor)
 
-	fmt.Print(result)
-
 	// back to normal state
 	w.Close()
 	os.Stdout = old // restoring the real stdout
 	out := <-outC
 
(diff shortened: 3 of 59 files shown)


pkg/event/processor_test.go
```diff
@@ -449,3 +449,39 @@ func TestChanQueueEventProcessor_ProcessBatch(t *testing.T) {
...
+		WithEventDispatcher(NewMockDispatcher(100, false)))
+	processor.Start(exeCtx)
+
+	impression := BuildTestImpressionEvent()
+	conversion := BuildTestConversionEvent()
+
+	for i := 0; i < b.N; i++ {
+		processor.ProcessEvent(impression)
+		processor.ProcessEvent(conversion)
+		time.Sleep(1 * time.Millisecond)
```
mikecdavis: Doesn't adding the `time.Sleep` nullify the benchmark?

pkg/event/processor_test.go
```diff
@@ -449,3 +449,39 @@ func TestChanQueueEventProcessor_ProcessBatch(t *testing.T) {
 		assert.True(t, len(logEvent.Event.Visitors) >= 1)
 	}
 }
+
+func BenchmarkProcessor(b *testing.B) {
+	exeCtx := utils.NewCancelableExecutionCtx()
+	processor := NewBatchEventProcessor(
+		WithQueueSize(1000),
```
mikecdavis: Following the example in [this](https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go) article you can parameterize the benchmark.
```go
func benchmarkProcessor(qs int, b *testing.B) {
	exeCtx := utils.NewCancelableExecutionCtx()
	processor := NewBatchEventProcessor(
		WithQueueSize(qs)
		WithFlushInterval(10 * time.Millisecond),
		WithEventDispatcher(NewMockDispatcher(100, false)))
	processor.Start(exeCtx)

	impression := BuildTestImpressionEvent()
	conversion := BuildTestConversionEvent()

	for i := 0; i < b.N; i++ {
		processor.ProcessEvent(impression)
		processor.ProcessEvent(conversion)
		time.Sleep(500)
	}
	processor.Stop()
}

func BenchmarkQueueSize100(b *testing.B) { benchmarkProcessor(100, b) }
func BenchmarkQueueSize2000(b *testing.B) { benchmarkProcessor(2000, b) }
func BenchmarkQueueSize5000(b *testing.B) { benchmarkProcessor(5000, b) }
func BenchmarkQueueSize10000(b *testing.B) { benchmarkProcessor(10000, b) }
func BenchmarkQueueSize50000(b *testing.B) { benchmarkProcessor(50000, b) }
```

pkg/event/processor_test.go
```diff
@@ -449,3 +449,39 @@ func TestChanQueueEventProcessor_ProcessBatch(t *testing.T) {
...
+		time.Sleep(1 * time.Millisecond)
+	}
+}
+
+func BenchmarkProcessorDefault(b *testing.B) {
+	exeCtx := utils.NewCancelableExecutionCtx()
+	processor := NewBatchEventProcessor(
+		WithFlushInterval(10 * time.Millisecond),
+		WithEventDispatcher(NewMockDispatcher(100, false)))
+	processor.Start(exeCtx)
```
mikecdavis: Assuming we should `Stop()` as well?

pkg/event/processor_test.go
thomaszurkan-optimizely: There is literally no other way to let the go routines run that put in a sleep.  The sleep is now 500 nanoseconds.

pkg/event/processor_test.go
```diff
@@ -449,3 +448,153 @@ func TestChanQueueEventProcessor_ProcessBatch(t *testing.T) {
...
+func BenchmarkWithBatchSize(b *testing.B) {
+	logging.SetLogger(&NoOpLogger{})
+
+	merges := []struct {
+		name string
+		batchSize int
+		fun  func(bs int, b *testing.B) int
+	}{
+		{"BatchSize10", 10, benchmarkProcessorWithBatchSize},
+		{"BatchSize20", 20,benchmarkProcessorWithBatchSize},
```
mikecdavis: Nit, spacing

pkg/event/processor_test.go
```diff
@@ -449,3 +448,153 @@ func TestChanQueueEventProcessor_ProcessBatch(t *testing.T) {
...
+	benchmarkProcessorQueueSize(2000, b)
+}
+
+func benchmarkProcessor3000(b *testing.B) {
+	benchmarkProcessorQueueSize(3000, b)
+}
+
+func benchmarkProcessor4000(b *testing.B) {
+	benchmarkProcessorQueueSize(4000, b)
+}
```
mikecdavis: Given how you are parameterizing within `BenchmarkWithQueueSize` you can get rid of these functions.

pkg/event/processor_test.go
```diff
@@ -449,3 +448,153 @@ func TestChanQueueEventProcessor_ProcessBatch(t *testing.T) {
...
+
+	conversion := BuildTestConversionEvent()
+
+	for i := 0; i < b.N; i++ {
+		processor.ProcessEvent(conversion)
+	}
+
+	exeCtx.TerminateAndWait()
+
+	return dispatcher.Events.Size()
```
mikecdavis: Are these `return` values use for anything?

pkg/logging/logger.go
```diff
@@ -46,17 +48,23 @@ const (
 )
 
 func init() {
+	mutex.Lock()
 	defaultLogConsumer = NewFilteredLevelLogConsumer(LogLevelInfo, os.Stdout)
+	mutex.Unlock()
```
mikecdavis: I know this was already there, but is this `init()` needed or is it preferable to set during definition?
```go
var defaultLogConsumer OptimizelyLogConsumer = NewFilteredLevelLogConsumer(LogLevelInfo, os.Stdout)
```

pkg/logging/logger.go
```diff
@@ -46,17 +48,23 @@ const (
...
 }
 
 // SetLogger replaces the default logger with the given logger
 func SetLogger(logger OptimizelyLogConsumer) {
-	defaultLogConsumer = logger
+	mutex.Lock()
+	 defaultLogConsumer = logger
```
mikecdavis: Nit, spacing.

pkg/logging/logger.go
```diff
@@ -30,6 +31,7 @@ func (l LogLevel) String() string {
 }
 
 var defaultLogConsumer OptimizelyLogConsumer
+var mutex = &sync.Mutex{}
```
mikecdavis: Curious what issue was seen that we needed to protect the logger. The operations below seem to be atomic and shouldn't require a mutex, but I might be missing something.

pkg/event/processor_test.go
```diff
@@ -449,3 +448,153 @@ func TestChanQueueEventProcessor_ProcessBatch(t *testing.T) {
...
+func (l *NoOpLogger) Log(level logging.LogLevel, message string, fields map[string]interface{}) {
+}
+
+
+func (l *NoOpLogger) SetLogLevel(level logging.LogLevel) {
+
+}
+
+func BenchmarkWithQueueSize(b *testing.B) {
+	logging.SetLogger(&NoOpLogger{})
```
mikecdavis: Can you elaborate on why we needed to set a `NoOpLogger`?

pkg/logging/logger.go
thomaszurkan-optimizely: since there is a SetLogger.  I actually think this is safer.

pkg/logging/logger.go
thomaszurkan-optimizely: so, setLogger can be called at anytime.  During benchmarking, I was calling set at the begging and resetting at the end.  This caused a race condition which is now resolved.

pkg/event/processor_test.go
thomaszurkan-optimizely: yes, they are now used in the queue size tests to make sure that all events are processed.

pkg/event/processor.go
```diff
@@ -33,7 +33,7 @@ import (
 
 // Processor processes events
 type Processor interface {
-	ProcessEvent(event UserEvent)
+	ProcessEvent(event UserEvent) bool
```
mikeng13: Is the bool return value ever used? I don’t see why this is necessary unless the client makes use of it

pkg/event/processor_test.go
thomaszurkan-optimizely: Added stop via context terminate.

pkg/event/processor.go
mikeng13: 👍

go.mod
```diff
@@ -3,13 +3,17 @@ module github.com/optimizely/go-sdk
 go 1.12
 
 require (
+	github.com/DATA-DOG/godog v0.7.13
```
mikeng13: Can you remove these please? They are used only for FSC testing, which shoudn't be included in our main deps

PR #189, opened 2019-11-18
feat(PollingConfigManager): Implemented caching headers in PollingConfigManager.

yasirfolio3:
1. Added support for caching datafiles using `Last-Modified`.

### DIFF
--- a/pkg/config/polling_manager.go
+++ b/pkg/config/polling_manager.go
@@ -19,6 +19,7 @@ package config
 
 import (
 	"fmt"
+	"net/http"
 	"sync"
 	"time"
 
@@ -36,6 +37,12 @@ const DefaultPollingInterval = 5 * time.Minute // default to 5 minutes for polli
 // DatafileURLTemplate is used to construct the endpoint for retrieving the datafile from the CDN
 const DatafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json"
 
+// ModifiedSince header key for request
+const ModifiedSince = "If-Modified-Since"
+
+// LastModified header key for response
+const LastModified = "Last-Modified"
+
 var cmLogger = logging.GetLogger("PollingConfigManager")
 
 // PollingProjectConfigManager maintains a dynamic copy of the project config
@@ -44,6 +51,7 @@ type PollingProjectConfigManager struct {
 	pollingInterval    time.Duration
 	notificationCenter notification.Center
 	initDatafile       []byte
+	lastModified       string
 
 	configLock    sync.RWMutex
 	err           error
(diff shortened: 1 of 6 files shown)


pkg/config/polling_manager.go
```diff
@@ -89,21 +97,40 @@ func InitialDatafile(datafile []byte) OptionFunc {
...
 
 		if e != nil {
 			cmLogger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
 			cm.configLock.Lock()
 			closeMutex(e)
 			return
 		}
+
+		if code == 304 {
+			cmLogger.Debug("The datafile was not modified and won't be downloaded again")
```
pawels-optimizely: if we want to go this way, I am not sure if the logic : 
```
if projectConfig.GetRevision() == previousRevision {
		cmLogger.Debug(fmt.Sprintf("No datafile updates. Current revision number: %s", cm.projectConfig.GetRevision()))
		closeMutex(nil)
		return
	}
``` 
is still needed. It is saying the same thing. I am not even sure if there is a real case scenario for that logic to be executed.

msohailhussain: What about if anyone is providing URL template (server) which doesn't provide status 304? but keeps the datafile same?

pawels-optimizely: are you saying that somebody would try to reimplement requester ? Our default requester is using http libs and I think you will always get Last Modified header. @mikeng13 , what do you think ?

msohailhussain: no am saying, like in FSC we use our datafile server, if anyone using like that for testing their own stuff then it may cause loading everytime.

pkg/config/polling_manager.go
```diff
@@ -89,21 +97,40 @@ func InitialDatafile(datafile []byte) OptionFunc {
...
+		}
 
 		if e != nil {
 			cmLogger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
 			cm.configLock.Lock()
 			closeMutex(e)
 			return
 		}
+
+		if code == 304 {
```
pawels-optimizely: you can use http.StatusNotModified instead of 304 ( it will be clearer)

yasirfolio3: Updated all previously hardcoded status codes too.

msohailhussain:
@mikeng13 no we can't run FSC DFM right now. Steps definition to be added for that and also need to run datafile server (travis related change).

pkg/config/polling_manager.go
pawels-optimizely: looks good, tested the code with datafile changes.

PR #190, opened 2019-11-19
feat(gherkin-UPS): Implemented UPS support for gherkin.

yasirfolio3:
1. Added support for testing UPS on gherkin.
https://travis-ci.com/optimizely/go-sdk/jobs/258760864


msohailhussain:
@mikeng13 result of FSC

tests/integration/optlyplugins/save_error_ups.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/optimizely/go-sdk/pkg/decision"
+)
+
+// SaveErrorUserProfileService represents a user profile service with save error
+type SaveErrorUserProfileService struct {
+	NormalUserProfileService
+}
+
+// Save is used to save bucketing decisions for users
+func (s *SaveErrorUserProfileService) Save(userProfile decision.UserProfile) {
```
mikeng13: Shouldn't this error?

mikeng13:
@yasirfolio3 @msohailhussain  can you check why these tests won't run? https://travis-ci.com/optimizely/go-sdk/builds/137464741

tests/integration/support/client_wrapper.go
```diff
@@ -75,9 +77,20 @@ func GetInstance(datafileName string) *ClientWrapper {
...
-	decisionService := &optlyplugins.TestCompositeService{CompositeService: *decision.NewCompositeService("")}
 	eventProcessor.EventDispatcher = &optlyplugins.ProxyEventDispatcher{}
 
+	config, err := configManager.GetConfig()
+	if err != nil {
+		log.Fatal(err)
+	}
+	userProfileService := createUserProfileService(config, apiOptions)
+
+	experimentServiceOptions := []decision.CESOptionFunc{}
```
mikeng13: You can simplify these lines like:

```
compositeExperimentService := decision.NewCompositeExperimentService(
  decision.WithUserProfileService(userProfileService)
)
```

tests/integration/support/client_wrapper.go
```diff
@@ -75,9 +77,20 @@ func GetInstance(datafileName string) *ClientWrapper {
...
+	experimentServiceOptions = append(experimentServiceOptions, decision.WithUserProfileService(userProfileService))
+	compositeExperimentService := decision.NewCompositeExperimentService(experimentServiceOptions...)
+	compositeService := *decision.NewCompositeService("", decision.WithCompositeExperimentService(compositeExperimentService))
```
mikeng13: If there is an SDK key provided (not empty string), then we should use it, otherwise we won't be able to grab the notification center properly

tests/integration/optlyplugins/lookup_error_ups.go
```diff
@@ -0,0 +1,29 @@
+/****************************************************************************
```
mikeng13: Since we have so many of these let's just have a `user_profile_service` package in here

tests/integration/optlyplugins/normal_ups.go
```diff
@@ -0,0 +1,78 @@
...
+package optlyplugins
+
+import (
+	"sync"
+
+	"github.com/optimizely/go-sdk/pkg/decision"
+)
+
+// UPSHelper defines Helper methods for UPS
+type UPSHelper interface {
```
mikeng13: this can be in a `utils.go` file inside the `user_profile_service` package

tests/integration/support/client_wrapper.go
```diff
@@ -313,3 +327,42 @@ func (c *ClientWrapper) track(request models.APIOptions) (models.APIResponse, er
 	response.Result = "NULL"
 	return response, err
 }
+
+func createUserProfileService(config pkg.ProjectConfig, apiOptions models.APIOptions) decision.UserProfileService {
```
mikeng13: This doesn't need to be in the wrapper. We can make it helper method in `optlyplugins/user_profile_service` inside the `utils.go` file

tests/integration/support/steps.go
```diff
@@ -343,6 +365,67 @@ func (c *ScenarioCtx) PayloadsOfDispatchedEventsDontIncludeDecisions() error {
...
+	if err != nil {
+		return fmt.Errorf("Invalid Project Config")
+	}
+	parsedProfiles, err := parseYamlArray(value.Content, config)
+	if err != nil {
+		return fmt.Errorf("Invalid request for user profile service state")
+	}
+
+	var expectedProfiles []decision.UserProfile
+	for _, profile := range parsedProfiles {
```
mikeng13: This can be another util method inside of the `optlyplugins/user_profile_service` package

tests/integration/support/steps.go
```diff
@@ -343,6 +365,67 @@ func (c *ScenarioCtx) PayloadsOfDispatchedEventsDontIncludeDecisions() error {
...
+		expectedProfiles = append(expectedProfiles, userProfile)
+	}
+
+	actualProfiles := c.clientWrapper.UserProfileService.(optlyplugins.UPSHelper).GetUserProfiles()
+
+	success := false
+	for _, expectedProfile := range expectedProfiles {
+		success = false
+		for index, actualProfile := range actualProfiles {
+			if subset.Check(expectedProfile, actualProfile) {
```
mikeng13: Should this be a subset check? SHouldn't we check for strict equality between the actualProfiles and the expectedProfiles?

tests/integration/support/client_wrapper.go
yasirfolio3: We haven't yet implemented support for DFM tests, once we do, this will be updated with the provided `sdkKey`.

tests/integration/support/steps.go
yasirfolio3: This is because during conversion from map to array during parsing, the order of elements is not always accurate, because of which, direct subset comparison randomly fails. I have now implemented a check to validate the length of both arrays aswell.

yasirfolio3: Currently, FSC also checks for subset comparison:
`Then("the User Profile Service state should be", function(expectedState) {
  // TODO: allow keys in this state
  expectedState = this.load(expectedState);
  expect(this.context.user_profiles).to.containSubset(expectedState);
});`
 i tried it with equality comparison but some of the tests because of greater number of profiles returned by the sdk.

tests/integration/support/client_wrapper.go
mikeng13: ok can you just add a TODO here about it

tests/integration/support/steps.go
mikeng13: Isn't it a problem that we are returning a greater number of profiles than we expected with just the calls we have?

tests/integration/support/steps.go
```diff
@@ -343,6 +365,51 @@ func (c *ScenarioCtx) PayloadsOfDispatchedEventsDontIncludeDecisions() error {
...
+		}
+		if !success {
+			break
+		}
+	}
+	if success {
+		return nil
+	}
+
+	return fmt.Errorf("User profile state not equal")
```
mikeng13: We should log the different found between the user profiles

tests/integration/support/steps.go
```diff
@@ -343,6 +365,51 @@ func (c *ScenarioCtx) PayloadsOfDispatchedEventsDontIncludeDecisions() error {
...
+}
+
+// ThereIsNoUserProfileState checks that UPS is empty
+func (c *ScenarioCtx) ThereIsNoUserProfileState() error {
+	actualProfiles := c.clientWrapper.UserProfileService.(userprofileservice.UPSHelper).GetUserProfiles()
+	if len(actualProfiles) == 0 {
+		return nil
+	}
+	return fmt.Errorf("User profile state not empty")
```
mikeng13: Also log what the user profile was populated with

tests/integration/support/steps.go
oakbani: @mikeng13 I think we do that in FSC to logically group the tests in Scenario Outlines. Each example test can result in a different user profile state at the end. So for equality check we will have to separate them out in multiple scenarios. 

I suggest that we modify the wording of this comparison. And write some tests based on equality.

msohailhussain: Responded on slack.

tests/integration/support/steps.go
yasirfolio3: done 👍

PR #194, opened 2019-11-22
ci(nl): enable notification listener

yasirfolio3:
# Summary
- Enabled notification-listener tests.
- Created ExperimentFeatureMap in projectConfig to be used to check if experimentId belongs to a feature which is required to send notification with type `feature-test`.
- Fixed an issue in `CompositeService` where notification was not being sent if variation was `nil`.
- Fixed an issue in `ToFloat` utils method which was causing crashes.
- Fixed an issue in `CompositeService` where featureKey was being sent as `experimentKey ` in decision notifications.
- Updated decisionNotification callback implementation for gherkin.


pkg/config/datafileprojectconfig/mappers/feature.go
```diff
@@ -40,6 +41,7 @@ func MapFeatures(featureFlags []datafileEntities.FeatureFlag, rolloutMap map[str
 			if experiment, ok := experimentMap[experimentID]; ok {
 				featureExperiments = append(featureExperiments, experiment)
 			}
+			experimentFeatureMap[experimentID] = append(experimentFeatureMap[experimentID], featureFlag.ID)
```
mikeng13: Instead of doing this, we can simplify and add an extra bool property to `entities.Experiment` called `IsFeatureExperiment` so then later on we can just check for this property instead of having to compute it from the ProjectConfig.

pkg/decision/evaluator/matchers/utils/to_float.go
```diff
@@ -5,14 +5,18 @@ import "reflect"
 
 // ToFloat attempts to convert the given value to a float
 func ToFloat(value interface{}) (float64, bool) {
+
+	if value == nil {
```
mikeng13: nice catch here! Can you please add a unit test for it?

tests/integration/optlyplugins/test_composite_service.go
```diff
@@ -68,15 +71,17 @@ func (c *TestCompositeService) decisionNotificationCallback(notification notific
 	c.listenersCalled = append(c.listenersCalled, model)
 }
 
-func getDecisionInfoForNotification(notification notification.DecisionNotification) map[string]interface{} {
+func getDecisionInfoForNotification(notify notification.DecisionNotification) map[string]interface{} {
```
mikeng13: can call this `decisionNotification` instead

tests/integration/optlyplugins/test_composite_service.go
```diff
@@ -49,7 +50,9 @@ func (c *TestCompositeService) AddListeners(listeners map[string]int) {
 
 // GetListenersCalled - Returns listeners called
 func (c *TestCompositeService) GetListenersCalled() []models.DecisionListener {
-	return c.listenersCalled
+	listenerCalled := c.listenersCalled
+	c.listenersCalled = nil
```
mikeng13: Why are we clearing it out here? Can you put a docstring to explain?

pkg/config/datafileprojectconfig/mappers/feature_test.go
```diff
@@ -43,13 +43,16 @@ func TestMapFeatures(t *testing.T) {
 	rolloutMap := map[string]entities.Rollout{
 		"41111": rollout,
 	}
-	experiment31111 := entities.Experiment{ID: "31111"}
-	experiment31112 := entities.Experiment{ID: "31112"}
+	experiment31111 := entities.Experiment{ID: "31111", IsFeatureExperiment: false}
```
mikeng13: nit:  the zero-value for `IsFeatureExperiment` should already be false so there is no need to set it explicitly.

PR #195, opened 2019-11-25
refac: put URL as a parameter for a requester method

pawels-optimizely:


pkg/utils/requester.go
```diff
@@ -33,16 +33,19 @@ import (
 
 const defaultTTL = 5 * time.Second
 
+// DatafileAPI is used as a default API for retrieving the datafile from the CDN
+const DatafileAPI = "https://cdn.optimizely.com/datafiles"
```
mikeng13: This is wrong, we should pass this into the `Get`, `Post`, etc because only the polling manager uses this template. We should look at the requester as just a generic client that can make http requests to whatever URI we give it

pkg/config/polling_manager.go
```diff
@@ -34,9 +34,6 @@ import (
 // DefaultPollingInterval sets default interval for polling manager
 const DefaultPollingInterval = 5 * time.Minute // default to 5 minutes for polling
 
-// DatafileURLTemplate is used to construct the endpoint for retrieving the datafile from the CDN
-const DatafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json"
```
mikeng13: This should stay here and be used to construct the URL of the request we make with the requester.

pkg/utils/requester.go
```diff
@@ -76,18 +86,18 @@ func Headers(headers ...Header) func(r *HTTPRequester) {
...
 
 // NewHTTPRequester makes Requester with api and parameters. Sets defaults
-// url has a complete url of the request like https://cdn.optimizely.com/datafiles/24234.json
-func NewHTTPRequester(url string, params ...func(*HTTPRequester)) *HTTPRequester {
+// api has the base part of request's url, like http://localhost/api/v1
+func NewHTTPRequester(params ...func(*HTTPRequester)) *HTTPRequester {
 
 	res := HTTPRequester{
-		url:     url,
+		api:     DatafileAPI,
```
mikeng13: I don't think we should be specifying the APIs for the requester like this. The requester should remain agnostic of what the API or base URL is. We should let it just take any URL and send the http request to it.

pkg/utils/requester.go
pawels-optimizely: addressed

pkg/client/factory_test.go
```diff
@@ -75,15 +71,8 @@ func TestClientWithPollingConfigManager(t *testing.T) {
 
 func TestClientWithPollingConfigManagerRequester(t *testing.T) {
 
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
mikeng13: Won't this make it so that the polling config manager starts polling for real? Seems like it could be a lot of noise for the unit tests as we'll start seeing 403s or 404s?

pkg/config/polling_manager.go
```diff
@@ -61,13 +62,11 @@ type PollingProjectConfigManager struct {
 // OptionFunc is a type to a proper func
 type OptionFunc func(*PollingProjectConfigManager)
 
-// DefaultRequester is an optional function, sets default requester based on a key.
-func DefaultRequester(sdkKey string) OptionFunc {
+// DefaultRequester is an optional function, sets default requester
+func DefaultRequester() OptionFunc {
```
mikeng13: I am not sure we need this OptionFunc as it is never used. We always instantiate a HTTPRequester in the `NewPollingProjectConfigManager` method.

pkg/event/dispatcher.go
```diff
@@ -46,8 +46,8 @@ type HTTPEventDispatcher struct {
 // DispatchEvent dispatches event with callback
 func (*HTTPEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
 
-	requester := utils.NewHTTPRequester(event.EndPoint)
-	_, _, code, err := requester.Post(event.Event)
+	requester := utils.NewHTTPRequester()
```
mikeng13: Are you planning on tackling the requester re-use in a different PR?

pkg/utils/requester.go
```diff
@@ -100,40 +98,40 @@ func NewHTTPRequester(url string, params ...func(*HTTPRequester)) *HTTPRequester
 }
 
 // Get executes HTTP GET with url and optional extra headers, returns body in []bytes
-// url created as url+sdkKey.json
-func (r HTTPRequester) Get(headers ...Header) (response []byte, responseHeaders http.Header, code int, err error) {
-	return r.Do("GET", nil, headers)
+// url created as api+sdkKey.json
```
mikeng13: This comment about `api+sdkKey.json` is very specific to the datafile. The requester shouldn't care about this detail. We can just remove this part of the comment.

pkg/config/polling_manager.go
```diff
@@ -79,6 +78,13 @@ func Requester(requester utils.Requester) OptionFunc {
 	}
 }
 
+// DatafileTemplate is an optional function, sets a passed datafile URL template
```
mikeng13: `DatafileURLTemplate`

pawels-optimizely: const DatafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json", go does not allow to use the same name for a var and a function

pkg/event/dispatcher.go
pawels-optimizely: yes

pkg/config/polling_manager.go
pawels-optimizely: removed

pkg/client/factory_test.go
pawels-optimizely: addressed

PR #196, opened 2019-11-26
chore(docs): Clean up public doc strings on client, factory and top-level components.

mikeng13:
# Summary
- Clean up public doc strings on client, factory and top-level components.
- Makes private some of the methods on the `BatchEventProcessor` which do not need to be public.


pkg/config/polling_manager.go
```diff
@@ -58,7 +59,7 @@ type PollingProjectConfigManager struct {
 	projectConfig pkg.ProjectConfig
 }
 
-// OptionFunc is a type to a proper func
+// OptionFunc is s used to provide custom configuration to the PollingProjectConfigManager.
```
thomaszurkan-optimizely: is s used...?

PR #198, opened 2019-11-26
feat(onTrack): Added onTrack callback

msohailhussain:
When track is called, onTrack should be used to notify user if definition for onTrack is set in the client.


msohailhussain:
Sure, let me revise. I will discuss with you first.

pkg/client/client.go
```diff
@@ -332,7 +333,16 @@ func (o *OptimizelyClient) Track(eventKey string, userContext entities.UserConte
 	}
 
 	userEvent := event.CreateConversionUserEvent(projectConfig, configEvent, userContext, eventTags)
-	o.EventProcessor.ProcessEvent(userEvent)
+	if o.EventProcessor.ProcessEvent(userEvent) && o.NotificationCenter != nil {
+		trackNotification := notification.TrackNotification{Type: notification.Track, EventKey: eventKey, UserContext: userContext, EventTags: eventTags}
+		var payload []interface{}
+		payload = append(payload, trackNotification, userEvent)
```
mikeng13: Why make this an array? We can add the UserEvent to the track notification struct

pkg/client/client.go
```diff
@@ -431,6 +441,45 @@ func (o *OptimizelyClient) getExperimentDecision(experimentKey string, userConte
...
 }
 
+// OnTrack registers a handler for Track notifications
+func (o *OptimizelyClient) OnTrack(callback func(notification notification.TrackNotification, userEvent event.UserEvent)) (int, error) {
+	if o.NotificationCenter == nil {
+		return 0, fmt.Errorf("no notification center found")
+	}
+	handler := func(payload interface{}) {
+		success := false
+		if trackPayload, ok := payload.([]interface{}); ok {
```
mikeng13: Should be casting to a TrackNotification directly

msohailhussain:
discussed offline, keeping it on hold. TrackNotification will have cyclic import issue if `UserEvent` is added there.

pkg/client/client_test.go
```diff
@@ -107,9 +111,94 @@ func (TestConfig) GetClientVersion() string {
 	return "1.0.0"
 }
 
+func TestAddandRemoveOnTrack(t *testing.T) {
```
mikeng13: Can you create a test suite for these new tests to have better code reuse? 

We already have some test suites towards the end like this:

func TestClientTestSuite(t *testing.T) {
	suite.Run(t, new(ClientTestSuiteAB))
	suite.Run(t, new(ClientTestSuiteFM))
}

So you can add 1 test suite for the notifications APIs and another one for event tracking for the tests you added further below. Don't worry about refactoring the current track tests, you can simply add the track notification track tests in the test suite

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,344 @@ func TestClose(t *testing.T) {
...
+
+	isTrackCalled := false
+	onTrack := func(eventKey string, userContext entities.UserContext, eventTags map[string]interface{}, conversionEvent event.ConversionEvent) {
+		isTrackCalled = true
+		s.Equal("sample_conversion", eventKey)
+		s.Equal(expectedUserContext, userContext)
+		s.Equal(*s.mockProcessor.Events[0].Conversion, conversionEvent)
+	}
+
+	client := OptimizelyClient{
```
mikeng13: This client initialization is repeated throughout the tests in the suite. We can initialize in the test setup

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,344 @@ func TestClose(t *testing.T) {
...
+	onTrack := func(eventKey string, userContext entities.UserContext, eventTags map[string]interface{}, conversionEvent event.ConversionEvent) {
+		isTrackCalled = true
+	}
+
+	client := OptimizelyClient{
+		ConfigManager:   ValidProjectConfigManager(),
+		DecisionService: s.mockDecisionService,
+		EventProcessor:  s.mockProcessor,
+	}
+	client.OnTrack(onTrack)
```
mikeng13: This is incorrect, we need to populate the notification center to ensure that there was nothing router through it.

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,344 @@ func TestClose(t *testing.T) {
...
+
+	expectedUserContext := entities.UserContext{ID: "1212121", Attributes: map[string]interface{}{}}
+
+	isTrackCalled := false
+	onTrack := func(eventKey string, userContext entities.UserContext, eventTags map[string]interface{}, conversionEvent event.ConversionEvent) {
+		isTrackCalled = true
+	}
+
+	mockNotificationCenter := new(MockNotificationCenter)
+	mockNotificationCenter.On("Send", mock.Anything, mock.Anything).Return(fmt.Errorf(""))
```
mikeng13: Let's not accept anything. We should correctly type the arguments received in the send method. We don't have to do it for AddHandler below we cannot compare functions. HOwever we should at least ensure that the argument is of type func

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,344 @@ func TestClose(t *testing.T) {
...
+	s.Error(err)
+
+	err = client.RemoveOnTrack(0)
+	s.Error(err)
+}
+
+func (s *ClientTestSuiteTrackNotification) TestOnTrackThrowsErrorWhenAddHandlerFails() {
+
+	mockNotificationCenter := new(MockNotificationCenter)
+	mockNotificationCenter.On("AddHandler", mock.Anything, mock.Anything).Return(-1, fmt.Errorf(""))
```
mikeng13: Assert the type, don't use mock.Anything

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,344 @@ func TestClose(t *testing.T) {
...
+	}
+	id, err := client.OnTrack(onTrack)
+	s.Equal(0, id)
+	s.Error(err)
+}
+
+func (s *ClientTestSuiteTrackNotification) TestRemoveOnTrackThrowsErrorWhenRemoveHandlerFails() {
+
+	mockNotificationCenter := new(MockNotificationCenter)
+	mockNotificationCenter.On("AddHandler", mock.Anything, mock.Anything).Return(1, nil)
```
mikeng13: Same as above.

pkg/client/factory.go
```diff
@@ -64,6 +65,7 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 	}
 
 	appClient := &OptimizelyClient{executionCtx: executionCtx}
+	appClient.NotificationCenter = registry.GetNotificationCenter(f.SDKKey)
```
mikeng13: This doesn't need to be set after the object init. You can put it right after the exec context

pkg/client/client_test.go
```diff
@@ -1950,17 +1950,25 @@ type ClientTestSuiteTrackEvent struct {
 	suite.Suite
 	mockProcessor       *MockProcessor
 	mockDecisionService *MockDecisionService
+	client              OptimizelyClient
```
msohailhussain: Client shouldn't be part of this TestSuite. Only mocks should come here..

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+		ConfigManager:      ValidProjectConfigManager(),
+		DecisionService:    s.mockDecisionService,
+		EventProcessor:     s.mockProcessor,
+		NotificationCenter: notification.NewNotificationCenter(),
+	}
+}
+
+func (s *ClientTestSuiteTrackEvent) TestTrackWithNotification() {
+
+	s.mockProcessor.On("ProcessEvent", mock.Anything).Return(true)
```
msohailhussain: `mock.Anything` is very general. We are using ConversionEvent, so please use that type.

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+
+	isTrackCalled := false
+	onTrack := func(eventKey string, userContext entities.UserContext, eventTags map[string]interface{}, conversionEvent event.ConversionEvent) {
+		isTrackCalled = true
+		s.Equal("sample_conversion", eventKey)
+		s.Equal(expectedUserContext, userContext)
+		s.Equal(*s.mockProcessor.Events[0].Conversion, conversionEvent)
+	}
+
+	s.client.OnTrack(onTrack)
```
msohailhussain: add return ID and assert that.

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+	s.NoError(err)
+	s.True(isTrackCalled)
+	s.Equal(1, len(s.mockProcessor.Events))
+	s.Equal("1212121", s.mockProcessor.Events[0].VisitorID)
+	s.Equal("15389410617", s.mockProcessor.Events[0].EventContext.ProjectID)
+}
+
+func (s *ClientTestSuiteTrackEvent) TestTrackWithNotificationAndEventTag() {
+
+	s.mockProcessor.On("ProcessEvent", mock.Anything).Return(true)
```
msohailhussain: same as above.

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+
+	s.NoError(err)
+	s.True(isTrackCalled)
+	s.Equal(1, len(s.mockProcessor.Events))
+	s.Equal("1212121", s.mockProcessor.Events[0].VisitorID)
+	s.Equal("15389410617", s.mockProcessor.Events[0].EventContext.ProjectID)
+}
+
+func (s *ClientTestSuiteTrackEvent) TestTrackNotificationNotCalledWhenEventProcessorReturnsFalse() {
+	s.mockProcessor.On("ProcessEvent", mock.Anything).Return(false)
```
msohailhussain: same as above and do it in all other places.

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+	isTrackCalled := false
+	onTrack := func(eventKey string, userContext entities.UserContext, eventTags map[string]interface{}, conversionEvent event.ConversionEvent) {
+		isTrackCalled = true
+	}
+
+	s.client.OnTrack(onTrack)
+	err := s.client.Track("sample_conversion", entities.UserContext{ID: "1212121", Attributes: map[string]interface{}{}}, map[string]interface{}{})
+
+	s.NoError(err)
+	s.Equal(1, len(s.mockProcessor.Events))
```
msohailhussain: How it can be 1. I think this is confusing. The best way would be to mock event.CreateConversionEvent if possible. and return any invalid type and let ProcessEvent do its job. It will return error while invalid type is passed.

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+
+func (s *ClientTestSuiteTrackEvent) TestTrackNotificationNotCalledWhenInvalidParamsProvided() {
+
+	isTrackCalled := false
+	onTrack := func(eventKey string, userContext entities.UserContext, eventTags map[string]interface{}, conversionEvent event.ConversionEvent) {
+		isTrackCalled = true
+	}
+
+	s.client.OnTrack(onTrack)
+	err := s.client.Track("bob", entities.UserContext{ID: "1212121", Attributes: map[string]interface{}{}}, map[string]interface{}{})
```
msohailhussain: this is confusing, what's invalid here. Please add comment.

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+	s.mockDecisionService = new(MockDecisionService)
+	s.client = OptimizelyClient{
+		ConfigManager:      ValidProjectConfigManager(),
+		DecisionService:    s.mockDecisionService,
+		EventProcessor:     s.mockProcessor,
+		NotificationCenter: notification.NewNotificationCenter(),
+	}
+}
+
+func (s *ClientTestSuiteTrackNotification) TestAddandRemoveOnTrack() {
```
msohailhussain: Please break down this test into 2 or 3 tests. 
- Add multiple ontrack and assert
- Add multiple ontrack, remove by ID and assert then. 
- Add, remove, add and assert. 
Make sure, IDs are always increased(unique).

pkg/client/client_test.go
```diff
@@ -1910,7 +1946,303 @@ func TestClose(t *testing.T) {
...
+	for _, id := range callbackIds {
+		err = s.client.RemoveOnTrack(id)
+		s.NoError(err)
+	}
+	err = s.client.Track("sample_conversion", userContext, map[string]interface{}{})
+	s.NoError(err)
+	s.Equal(0, numberOfCalls)
+}
+
+func (s *ClientTestSuiteTrackNotification) TestOnTrackThrowsErrorWithoutNotificationCenter() {
```
msohailhussain: separate into two tests.

pkg/client/client_test.go
yasirfolio3: Updated the test name.

msohailhussain:
LGTM.

PR #199, opened 2019-11-27
refac: Use a single HTTP client throughout the HTTPEventDispatcher life-cycle

pawels-optimizely:


pkg/event/dispatcher.go
```diff
@@ -41,13 +41,13 @@ type Dispatcher interface {
 
 // HTTPEventDispatcher is the HTTP implementation of the Dispatcher interface
 type HTTPEventDispatcher struct {
+	requester *utils.HTTPRequester
 }
 
 // DispatchEvent dispatches event with callback
-func (*HTTPEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
+func (eventDispacher *HTTPEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
```
mikeng13: nit: should this be shortened to something like `ed` like we do for other func receivers?

pkg/event/processor_test.go
```diff
@@ -506,7 +504,7 @@ func BenchmarkWithQueueSize(b *testing.B) {
 	for _, merge := range merges {
 		var totalSent = 0
 		var numberRun = 0
-		b.Run(merge.name, func (b *testing.B) {
+		b.Run(merge.name, func(b *testing.B) {
```
mikeng13: I think this should've been caught by our linters. Can we update them to catch these in a follow-up PR?

pawels-optimizely: this is IDE's thing,  there is difference between Goland and VSC, I will look into that

pkg/event/dispatcher.go
pawels-optimizely: addressed

PR #200, opened 2019-11-28
feat(gherkin-FV): Implemented forced-variation for gherkin.

yasirfolio3:
# Summary
- Added support for forced-variation in gherkin.
- Created VariationKeyMap in ExperimentStruct which was required throughout the SDK.

# Test Result
https://travis-ci.com/optimizely/go-sdk/jobs/263744085


pkg/decision/experiment_override_service.go
```diff
@@ -62,10 +81,31 @@ func (m *MapExperimentOverridesStore) GetVariation(overrideKey ExperimentOverrid
...
+			return false
+		}
+		if variationID, ok := experiment.VariationKeyToIDMap[variationKey]; ok {
+			if _, ok := experiment.Variations[variationID]; ok {
+				assignVariationKey()
+				return true
+			}
+		}
+	}
+	return false
```
mjc1283: @msohailhussain @yasirfolio3 Could you explain why you are making this change?

pkg/decision/experiment_override_service.go
```diff
@@ -62,10 +81,31 @@ func (m *MapExperimentOverridesStore) GetVariation(overrideKey ExperimentOverrid
...
+		m.overridesMap[overrideKey] = variationKey
+		m.mutex.Unlock()
+	}
+	// Assign directly if no config provided
+	if m.config == nil {
+		assignVariationKey()
+		return true
+	}
+	// Check if experiment and variation exist
+	if experiment, err := m.config.GetExperimentByKey(overrideKey.ExperimentKey); err == nil {
```
mikeng13: Hmm I am not sure this is a valid thing to do. The project config is not static. It can change at any time and this reference would be stale. This experiment/variation validation is already being done in `GetDecision` so we don't really have to have this check during the setting of the override

yasirfolio3: One of the tests in fsc sets a valid variation first, then tries setting an invalid variation and expects variation to be not set. The expected result is the first variation set. This test fails with the previous implementation if we dont check for valid variation during setVariation method.

mikeng13: @mjc1283 what do you think? This is a bit cumbersome especially if we do check it at the bottom, and granted that it's a QA tool, I don't think we should over-engineer

mikeng13:
Can you also include a link to passing FSC tests with forced variations enabled?

yasirfolio3:
> Can you also include a link to passing FSC tests with forced variations enabled?

@mikeng13 Updated tests results in the summary.
We need to add these tags for testing:
`~@OPTIMIZELY_CONFIG&&~@INVALID_FORCED_VARIATION&&~@LOG_EVENT_LISTENER`
This FSC also needs to be merged first:
https://github.com/optimizely/fullstack-sdk-compatibility-suite/pull/365

mjc1283:
@yasirfolio3 @mikeng13 @msohailhussain I think the changes in this PR look good. But as discussed in standup, I suggest we can delete the FSC test that tests validation in set forced variation. Can we do this instead of introducing a tag?

yasirfolio3:
> @yasirfolio3 @mikeng13 @msohailhussain I think the changes in this PR look good. But as discussed in standup, I suggest we can delete the FSC test that tests validation in set forced variation. Can we do this instead of introducing a tag?

yep, i have refactored the fsc PR 👍

msohailhussain:
https://travis-ci.com/optimizely/go-sdk/jobs/264187892

PR #201, opened 2019-12-02
feat(gherkin-DFM): Implemented DFM for gherkin.

yasirfolio3:
# Summary
- Added support for DFM in gherkin.
- Made a minor change to `PollingProjectConfigManager` to support adding `projectConfigUpdateHandlers` during initialization for test purposes.


yasirfolio3:
@mikeng13 Passing tests can be seen here for `DFM`:
https://travis-ci.com/optimizely/go-sdk/jobs/262046142

scripts/setup-integration-tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+. ~/.nvm/nvm.sh
+mkdir -p $FSC_PATH
+pushd $FSC_PATH && git init && git fetch --depth=1 https://$CI_USER_TOKEN@github.com/optimizely/fullstack-sdk-compatibility-suite ${FSC_BRANCH:-master} && git checkout FETCH_HEAD
+ln -s features/support/datafiles/ public 
+pushd services/datafile && nvm install && nvm use && npm install && popd
+node services/datafile/ &> /dev/null &
+popd
```
msohailhussain: eol

pkg/config/polling_manager.go
```diff
@@ -50,13 +50,14 @@ var cmLogger = logging.GetLogger("PollingConfigManager")
...
-	requester           utils.Requester
-	sdkKey              string
+	datafileURLTemplate         string
+	initDatafile                []byte
+	lastModified                string
+	notificationCenter          notification.Center
+	pollingInterval             time.Duration
+	requester                   utils.Requester
+	sdkKey                      string
+	projectConfigUpdateHandlers []func(notification.ProjectConfigUpdateNotification)
```
mikeng13: @msohailhussain I mentioned to you offline how we should not pull in the handlers here, but instead externally call `OnProjectConfigUpdate` in a loop and attach the handlers that way.

msohailhussain: yes Mike, am working on it.

PR #204, opened 2019-12-03
(refact): Refactor processor benchmarking.

mikecdavis:
## Summary
* Create `CountingDispatcher` for asserting number of events dispatched.
* Refactored various benchmarks into a single benchmark with all permutations.

During the bug bash yesterday I noticed that the assertion that we were doing for the processor benchmarks were a bit too lax (e.g. `totalSent < numberRun`). I wanted to make that a more exact assertion so I ended up refactoring a bit. 

The resulting assertion looks more like `b.N != dispatcher.visitorCount` and is done with every benchmarking scenario. Running it on my machine there appears to be a one-off error with the ChanQueue implementation: (e.g.)

```
Total sent and run  1438934 1438935
FAIL: BenchmarkProcessor/Channel/BatchSize-10/QueueSize-10-8

Total sent and run  9999 10000
FAIL: BenchmarkProcessor/Channel/BatchSize-40/QueueSize-1000-8

Total sent and run  1719782 1719783
FAIL: BenchmarkProcessor/Channel/BatchSize-60/QueueSize-1000-8
```


mikecdavis:
Per multiple conversations, I'm going to just remove the `ChanQueue` entirely. We can revisit alternative implementations in the future, but now we have a descent framework for evaluation.

pkg/event/processor_test.go
```diff
@@ -467,119 +481,69 @@ func (l *NoOpLogger) SetLogLevel(level logging.LogLevel) {
...
+		{"InMemory", NewInMemoryQueue},
 	}
 
 	for _, merge := range merges {
-		var totalSent = 0
-		var numberRun = 0
-		b.Run(merge.name, func(b *testing.B) {
-			if numberRun == 0 {
-				numberRun = b.N
+		for i := 1.; i <= 5; i++ {
```
mikeng13: Nice! I like this

PR #206, opened 2019-12-03
refac: Clean up various factory methods to simplify initialization.

mikeng13:
# Summary
- Make initialization simpler
- Add option to pass in Event Dispatcher directly into factory


pkg/client/factory.go
```diff
@@ -77,18 +78,22 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 	if f.eventProcessor != nil {
 		appClient.EventProcessor = f.eventProcessor
 	} else {
-		appClient.EventProcessor = event.NewBatchEventProcessor(
+		var eventProcessorOptions = []event.BPOptionConfig{
```
pawels-optimizely: do we still need them: 
event.WithBatchSize(event.DefaultBatchSize),
			event.WithQueueSize(event.DefaultEventQueueSize),
			event.WithFlushInterval(event.DefaultEventFlushInterval),
since they use default values ?

mikeng13: You are right, we don't need em.

PR #207, opened 2019-12-04
test(eventprocessor): add a test for large batch size

thomaszurkan-optimizely:
## Summary
- An issue was brought up about the queue not flushing the right size when the queue size was 50.  I have added a unit test to prove that it works as expected.

## Test plan
- Unit test that shows given 100 events with a batch size of 50, that there are two dispatched events with 50 visitors each


mikeng13:
I added the template for the PR description, could you please fill it out?

PR #212, opened 2019-12-06
Refactor ExecutionCtx into ExecGroup.

mikecdavis:
## Summary
* Allow `OptimizelyFactory` to accept a plain old **context** object (POCO)
* Rework `ExecutionCtx` into `ExecGroup` (ala `errgroup`)
* Refactor `Start` methods to *NOT* manage their own goroutines, but assume to be executed within one (this is similar to `http.ListenAndServe`)

This refactor started when I wanted to manage the `OptimizelyClient` lifecycle in my application, but was surprised that I couldn't easily just pass in my own cancellable context. Instead I had to implement my own `ExecutableCtx`, which also had concepts I didn't feel were immediately relevant to what I was trying to accomplish. Digging further I saw how the `ExecutableCtx` was being used it followed more of the semantics of an [errgroup](https://godoc.org/golang.org/x/sync/errgroup) which means simply managing a "group" of goroutines.

An added side effect of this change is that `ExecGroup` is isolated to the client pkg and it's not leaked into the other implementations. Everything now just has to deal with `context` and overriding the default `context` feels much more idiomatic.

### DIFF
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -18,6 +18,7 @@
 package client
 
 import (
+	"context"
 	"errors"
 	"time"
 
@@ -34,10 +35,10 @@ type OptimizelyFactory struct {
 	Datafile []byte
 
 	configManager      config.ProjectConfigManager
+	ctx                context.Context
 	decisionService    decision.Service
 	eventDispatcher    event.Dispatcher
 	eventProcessor     event.Processor
-	executionCtx       utils.ExecutionCtx
 	userProfileService decision.UserProfileService
 	overrideStore      decision.ExperimentOverrideStore
 }
@@ -56,14 +57,15 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 		return nil, errors.New("unable to instantiate client: no project config manager, SDK key, or a Datafile provided")
 	}
 
-	var executionCtx utils.ExecutionCtx
-	if f.executionCtx != nil {
-		executionCtx = f.executionCtx
+	var ctx context.Context
+	if f.ctx != nil {
+		ctx = f.ctx
 	} else {
-		executionCtx = utils.NewCancelableExecutionCtx()
+		ctx = context.Background()
 	}
 
-	appClient := &OptimizelyClient{executionCtx: executionCtx, notificationCenter: registry.GetNotificationCenter(f.SDKKey)}
+	eg := utils.NewExecGroup(ctx)
+	appClient := &OptimizelyClient{execGroup: eg, notificationCenter: registry.GetNotificationCenter(f.SDKKey)}
 
 	if f.configManager != nil {
 		appClient.ConfigManager = f.configManager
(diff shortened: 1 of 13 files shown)


pkg/client/factory.go
```diff
@@ -103,11 +105,11 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 
 	// Initialize the default services with the execution context
 	if pollingConfigManager, ok := appClient.ConfigManager.(*config.PollingProjectConfigManager); ok {
-		pollingConfigManager.Start(appClient.executionCtx)
+		eg.Go(pollingConfigManager.Start)
```
mikeng13: I really like how we invert the relationship here

pkg/event/processor.go
```diff
@@ -146,12 +146,14 @@ func NewBatchEventProcessor(options ...BPOptionConfig) *BatchEventProcessor {
...
 	if p.EventDispatcher == nil {
-		p.EventDispatcher = NewQueueEventDispatcher(exeCtx.GetContext())
+		dispatcher := NewQueueEventDispatcher()
+		defer dispatcher.flushEvents()
+		p.EventDispatcher = dispatcher
 	}
 
-	p.startTicker(exeCtx)
+	p.startTicker(ctx)
 	pLogger.Debug("Batch event processor started")
```
mjc1283: With these changes, does this message actually only get logged when the processor is stopped? Since `startTicker` was changed to return only when the context signals done.

pkg/event/processor.go
```diff
@@ -146,12 +146,14 @@ func NewBatchEventProcessor(options ...BPOptionConfig) *BatchEventProcessor {
 }
 
 // Start initializes the event processor
-func (p *BatchEventProcessor) Start(exeCtx utils.ExecutionCtx) {
+func (p *BatchEventProcessor) Start(ctx context.Context) {
 	if p.EventDispatcher == nil {
-		p.EventDispatcher = NewQueueEventDispatcher(exeCtx.GetContext())
+		dispatcher := NewQueueEventDispatcher()
+		defer dispatcher.flushEvents()
```
mjc1283: Is this `flushEvents` necessary? `flushEvents` is called in `startTicker` when the context signals done.

pkg/client/factory_test.go
```diff
@@ -113,14 +115,23 @@ func TestClientWithDecisionServiceAndEventProcessorInOptions(t *testing.T) {
...
-	assert.Equal(t, client.executionCtx, testExecutionCtx)
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	client.execGroup.Go(func(ctx context.Context) {
+		<-ctx.Done()
+		wg.Done()
+	})
+
+	cancel()
```
pawels-optimizely: cool, we were missing that in the ExecutionContext implementation. The only awkward thing will happen when a user will define cancelable context and cancels with the client: 

```
ctx, _ := context.WithCancel(context.Background())
client.Close()
```
that will return :

 ```the cancel function returned by context.WithCancel should be called, not discarded, to avoid a context leak```

pkg/event/processor.go
mikecdavis: We are aggressively calling `flushEvents()` that's true. This particular change is to replace the defer in the `QueueEventDispatcher` "constructor". https://github.com/optimizely/go-sdk/pull/212/files#diff-663642fbf6b03877a468459e9e872cb5L142-L150

pkg/event/processor.go
mikecdavis: Good catch, I'll move the log line.

PR #216, opened 2019-12-12
feat: Implement OptimizelyConfig API

pawels-optimizely:
## Summary 

* Implemented OptimizelyConfig API for go sdk
* Performed real tests on datafile taken from swift tests.


pkg/client/optimizely_config.go
```diff
@@ -0,0 +1,163 @@
...
+	for _, feature := range features {
+
+		var optlyFeatureVariablesMap = map[string]OptimizelyVariable{}
+		for _, featureVarible := range feature.VariableMap {
+			optlyVariable := OptimizelyVariable{Key: featureVarible.Key, ID: featureVarible.ID,
+				Type: string(featureVarible.Type), Value: featureVarible.DefaultValue}
+			optlyFeatureVariablesMap[featureVarible.Key] = optlyVariable
+		}
+
+		optlyExperimentMap := getExperimentMap(features, feature.FeatureExperiments, variableByIDMap)
```
jaeopt: We do not need to rebuild map here. You can consider re-using experimentsMap built before and pick for each feature's experimentIds.

pkg/client/optimizely_config_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	optimizelyFactory := &OptimizelyFactory{}
+	s.optimizelyClient, err = optimizelyFactory.Client(WithConfigManager(projectConfigManager))
+	if err != nil {
+		s.Fail("unable to initialize optimizely client")
+	}
+
+}
+
+func (s *OptimizelyConfigTestSuite) TestOptlyConfig() {
```
jaeopt: add test case for returning null when config not ready

pkg/client/optimizely_config_test.go
```diff
@@ -0,0 +1,81 @@
...
+		s.Fail("error opening file:" + dataFileName)
+	}
+
+	projectConfigManager, e := config.NewStaticProjectConfigManagerFromPayload(datafile)
+	if e != nil {
+		s.Fail("error constructing config manager")
+	}
+
+	outputFileName := "testdata/optimizely_config_expected.json"
+	expectedOutput, er := ioutil.ReadFile(outputFileName)
```
jaeopt: I understand this file read and map guarantees that all data fields are public, so no need for extra access control validation. If it's intention, add comments to clarify it.

pawels-optimizely: comments provided

pkg/client/optimizely_config_test.go
pawels-optimizely: test case added

pkg/client/optimizely_config.go
pawels-optimizely: addressed

pkg/config/datafileprojectconfig/optimizely_config.go
```diff
@@ -0,0 +1,145 @@
...
+ *                                                                          *
+ * Unless required by applicable law or agreed to in writing, software      *
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package datafileprojectconfig //
+package datafileprojectconfig
```
mikeng13: This is the wrong package to put it in. The `OptimizelyConfig` shouldn't be tightly coupled to the concept of a `datafile`. If it's the package name confusion that we care about I'd rather us just do `pkg/config/optimizely_config.go` and have the linter ignore it heh. I understand that the package name kind of collides with the file and class names but that is more syntactically correct for us

pkg/client/client.go
```diff
@@ -486,6 +486,17 @@ func (o *OptimizelyClient) GetProjectConfig() (projectConfig config.ProjectConfi
 	return projectConfig, nil
 }
 
+// GetOptimizelyConfig returns OptimizelyConfig object
+func (o *OptimizelyClient) GetOptimizelyConfig() (optimizelyConfig *entities.OptimizelyConfig, err error) {
+
+	projectConfig, err := o.ConfigManager.GetConfig()
```
mikeng13: We can simply call `o.GetProjectConfig()` right? Also, let's make that method private since now everyone should be using this new one.

pkg/config/datafileprojectconfig/config.go
```diff
@@ -161,6 +169,16 @@ func (c DatafileProjectConfig) GetGroupByID(groupID string) (entities.Group, err
...
 }
 
+// GetOptimizelyConfig is the main interface to get OptimizelyConfig object
+func (c DatafileProjectConfig) GetOptimizelyConfig() *entities.OptimizelyConfig {
+
+	featuresList := c.GetFeatureList()
+	experimentsList := c.GetExperimentList()
+	revision := c.GetRevision()
+
+	return entities.NewOptimizelyConfig(featuresList, experimentsList, revision)
```
mikeng13: @jaeopt should we be caching this lazily instead of re-computing every time? This is tied directly to the project config so we don't have to worry about the cached version becoming stale.

pkg/config/datafileprojectconfig/config.go
```diff
@@ -127,6 +127,14 @@ func (c DatafileProjectConfig) GetFeatureList() (featureList []entities.Feature)
 	return featureList
 }
 
+// GetExperimentList returns an array of all the experiments
+func (c DatafileProjectConfig) GetExperimentList() (experimentList []entities.Experiment) {
```
mikeng13: Can you add a unit test for this method?

pkg/entities/optimizely_config_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+type OptimizelyConfigTestSuite struct {
+	suite.Suite
+	featureList              []Feature
+	experimentList           []Experiment
+	expectedOptimizelyConfig OptimizelyConfig
+}
+
+var featurelist = `[{"ID":"4482920077","Key":"mutex_group_feature","FeatureExperiments":[{"AudienceIds":[],"ID":"12198292373","LayerID":"12235440722","Key":"experiment_4000",
```
mikeng13: Why is this necessary? Is this the same contents from the json test files? Can't we just load those and reuse as json objects?

pkg/entities/testdata/optimizely_config_expected.json
```diff
@@ -0,0 +1,345 @@
...
+          "id":"2687470097",
+          "key":"s_foo",
+          "type":"string",
+          "value":"foo"
+        }
+      }
+    }
+  },
+  "revision":"9"
+}
```
mikeng13: nit: newline

pkg/config/datafileprojectconfig/test/optimizely_config_datafile.json
```diff
@@ -0,0 +1,268 @@
...
+      ],
+      "id": "12115595439"
+    }
+  ],
+  "attributes": [],
+  "botFiltering": false,
+  "accountId": "8272261422",
+  "events": [],
+  "revision": "9"
+}
```
mikeng13: nit: newline

pkg/config/datafileprojectconfig/test/optimizely_config_expected.json
```diff
@@ -0,0 +1,345 @@
...
+          "id":"2687470097",
+          "key":"s_foo",
+          "type":"string",
+          "value":"foo"
+        }
+      }
+    }
+  },
+  "revision":"9"
+}
```
mikeng13: nit: newline

pkg/entities/optimizely_config_test.go
pawels-optimizely: this is not exactly the same contents. These are just feature list and experiment list. This is used just to perform unit tests here (to keep the coverage). I am testing the whole thing again on project config level.

pkg/config/datafileprojectconfig/config.go
pawels-optimizely: my understanding is that we want to rebuild it every time for now since keeping that in memory will be expensive. But @jaeopt  can confirm.

pkg/config/datafileprojectconfig/config.go
pawels-optimizely: yes, I missed that

pkg/config/datafileprojectconfig/config.go
jaeopt: Yes we discussed the issue in a slack thread. We'll let client cache it for now and will visit again later to see the optimization is needed in the SDK level.

pkg/config/polling_manager.go
```diff
@@ -157,6 +158,7 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
 	}
 	cmLogger.Debug(fmt.Sprintf("New datafile set with revision: %s. Old revision: %s", projectConfig.GetRevision(), previousRevision))
 	cm.projectConfig = projectConfig
+	cm.optimizelyConfig = NewOptimizelyConfig(projectConfig)
```
mikeng13: We should do this lazily. So if `cm.optimizelyConfig != nil` then we create the optimizelyConfig

pkg/config/polling_manager.go
```diff
@@ -215,6 +217,13 @@ func (cm *PollingProjectConfigManager) GetConfig() (ProjectConfig, error) {
 	return cm.projectConfig, nil
 }
 
+// GetConfig returns the project config
+func (cm *PollingProjectConfigManager) GetOptimizelyConfig() *OptimizelyConfig {
+	cm.configLock.RLock()
+	defer cm.configLock.RUnlock()
+	return cm.optimizelyConfig
```
mikeng13: Here we lazily create the optimizelyConfig so that we only ever take up the memory if this method is used.

pkg/config/optimizely_config.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Package entities //
+package config
+
+import "github.com/optimizely/go-sdk/pkg/entities"
+
+// OptimizelyConfig is a snapshot of the experiments and features in the project config
+type OptimizelyConfig struct {
+	Revision       string
+	ExperimentsMap map[string]OptimizelyExperiment
```
mikeng13: Should these entities types be pointers? Does it matter if we return the pointer to the config?

pawels-optimizely: maps are already pointers to the underlying array already so it's ok to have it like that

pkg/config/polling_manager.go
pawels-optimizely: I will address that on the optimizely config

pkg/config/polling_manager.go
pawels-optimizely: this always returns either null or populated optimizely config. it always takes up the memory to construct it every time a datafile is updated.

pawels-optimizely: ok, we have clarified over IM

tests/integration/support/client_wrapper.go
```diff
@@ -120,6 +122,12 @@ func GetInstance(apiOptions models.APIOptions) *ClientWrapper {
 	return clientInstance
 }
 
+// GetProjectConfig gets the right config project
+func (c *ClientWrapper) GetProjectConfig() config.ProjectConfig {
+	return c.projectConfig
```
mikeng13: Instead of caching this, we can call `c.client.ConfigManager.GetConfig()`

pawels-optimizely: ok, addressed

pkg/config/polling_manager.go
```diff
@@ -215,6 +219,18 @@ func (cm *PollingProjectConfigManager) GetConfig() (ProjectConfig, error) {
...
 }
 
+// GetOptimizelyConfig returns the optimizely project config
+func (cm *PollingProjectConfigManager) GetOptimizelyConfig() *OptimizelyConfig {
+	cm.configLock.RLock()
+	defer cm.configLock.RUnlock()
+	if cm.optimizelyConfig != nil {
+		return cm.optimizelyConfig
+	}
+	optimizelyConfig := NewOptimizelyConfig(cm.projectConfig)
```
jaeopt: This is interesting - triggering the first OptConfig only when called by client. 
Tradeoff between memory saving for most clients who do not use OptConfig vs extra delay for 1st build.
It looks good as is, and we can consider alternate setting based on feedbacks.

PR #219, opened 2019-12-19
fix(notification): Remote notification only

msohailhussain:
### Summary
- Only downloaded datafiles should notify for OnConfigUpdate notification
- Hardcoded datafile should set `ProjectConfig` on initialization
- Remote datafile shouldn't block `NewPollingConfigManager`

### Test plan
- Added unit tests
- Revised few unit tests

### Issues
- Any application was using `sdk-key`, should have the code to block execution until `ProjectConfig` is available or register `OnProjectConfigUpdate` notification


pkg/config/polling_manager_test.go
```diff
@@ -44,6 +44,18 @@ func newExecGroup() *utils.ExecGroup {
 	return utils.NewExecGroup(context.Background())
 }
 
+// assertion method to wait for config or err for a specified period of time.
+func waitForConfigOrCancelTimeout(t *testing.T, configManager ProjectConfigManager, checkError bool) {
```
msohailhussain: I would like to hear if we can introduce `onReady` like other SDKs. Then this method is not needed.

pkg/config/polling_manager_test.go
```diff
@@ -192,6 +211,7 @@ func TestPollingGetOptimizelyConfig(t *testing.T) {
 	configManager.SyncConfig(mockDatafile2)
 	optimizelyConfig = configManager.GetOptimizelyConfig()
 	assert.Equal(t, "43", optimizelyConfig.Revision)
+	//mockRequester.AssertExpectations(t)
```
pawels-optimizely: @msohailhussain , not sure why I could not check for requester assert expectation. Can you take a look ?

pkg/config/polling_manager_test.go
msohailhussain: not needed. got mike's feedback.

pkg/config/polling_manager_test.go
msohailhussain: I don't know how it was working before. When SyncConfig is called with payload in the argument, it means no `Get` call should be asserted. I have removed explicit  `SyncConfig` call and now dependent on `Start` scheduled call. Very first call is called immediately, so no major wait in this unit test.

pkg/config/polling_manager.go
```diff
@@ -163,10 +164,10 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
 	}
 	closeMutex(nil)
 
-	if cm.notificationCenter != nil {
+	if cm.notificationCenter != nil && len(initDatafile) == 0 {
 		projectConfigUpdateNotification := notification.ProjectConfigUpdateNotification{
 			Type:     notification.ProjectConfigUpdate,
-			Revision: cm.projectConfig.GetRevision(),
+			Revision: projectConfig.GetRevision(),
```
msohailhussain: FYI. race condition fix

pkg/config/polling_manager_test.go
```diff
@@ -268,6 +302,77 @@ func TestNewPollingProjectConfigManagerOnDecision(t *testing.T) {
 	assert.Nil(t, err)
 }
 
+func TestNewPollingProjectConfigManagerHardcodedDatafile(t *testing.T) {
+	mockDatafile1 := []byte(`{"revision":"42"}`)
+	mockDatafile2 := []byte(`{"revision":"43"}`)
+	sdkKey := "test_sdk_key"
+
+	mockRequester := new(MockRequester)
```
mjc1283: This `mockRequester` isn't passed as an option to `NewPollingProjectConfigManager` via `WithRequester`, so the assertion on 316 isn't useful, right?

pkg/notification/manager.go
```diff
@@ -66,6 +72,8 @@ func (am *AtomicManager) Remove(id int) {
 
 // Send sends the notification to the registered handlers
 func (am *AtomicManager) Send(notification interface{}) {
+	am.lock.Lock()
```
mjc1283: If the user calls `Remove` in their notification callback, does that cause deadlock?

pkg/config/polling_manager_test.go
msohailhussain: Thanks @mjc1283 for pointing out, i have added requester there.

pkg/notification/manager.go
msohailhussain: How user will call `Remove` in the callback? I think it will be `race` exception?
I can try a sample code, and can check.

msohailhussain:
One more question here is, if user calls `Activate` or any other API before `projectconfig` initialization. What should be the behavior? Right now it returns error.

pkg/notification/manager.go
```diff
@@ -66,6 +72,8 @@ func (am *AtomicManager) Remove(id int) {
...
+	defer am.lock.Unlock()
 	for _, handler := range am.handlers {
 		handler(notification)
 	}
```
mikecdavis: Per our group discussion, we can first create an array of handlers, outside of the map, then trigger the handlers. That way we are not potentially modifying the Map while we're iterating.

msohailhussain:
Closing this PR since 
https://github.com/optimizely/go-sdk/pull/222
https://github.com/optimizely/go-sdk/pull/224
are already merged.

PR #220, opened 2019-12-31
fix: logging nil project config, adding more meaningful error messages

pawels-optimizely:
## Summary

Activate() returned vague runtime error (referencing null pointer).  After analysis and debugging, found 2 scenarios causing this:
- nil project config
- decision context is empty but variation is not nil somehow.

Added some protective logic against null references across client methods.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -20,6 +20,7 @@ package client
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"runtime/debug"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -477,6 +478,9 @@ func (o *OptimizelyClient) RemoveOnTrack(id int) error {
 
 func (o *OptimizelyClient) getProjectConfig() (projectConfig config.ProjectConfig, err error) {
 
+	if isNil(o.ConfigManager) {
+		return nil, errors.New("project config is not initialized")
```
mikeng13: Should be `project config manager is not initialized`

PR #222, opened 2020-01-02
fix(notification): race detected send notification issue

msohailhussain:
### Summary
- Fixed an issue when notification callbacks were calling Add/Remove of same notification manager. 
- All summary is in this PR. https://github.com/optimizely/go-sdk/pull/219

### TestPlan
- Added unit tests for race detected scenarios.

### DIFF
--- a/pkg/notification/manager.go
+++ b/pkg/notification/manager.go
@@ -81,13 +81,12 @@ func (am *AtomicManager) Send(notification interface{}) {
 	}
 }
 
-// Copy handlers and return it.
-func (am *AtomicManager) copyHandlers() map[uint32]func(interface{}) {
+// Return a copy of the given handlers
+func (am *AtomicManager) copyHandlers() (handlers []func(interface{})) {
 	am.lock.RLock()
 	defer am.lock.RUnlock()
-	m := make(map[uint32]func(interface{}), len(am.handlers))
-	for k, v := range am.handlers {
-		m[k] = v
+	for _, v := range am.handlers {
+		handlers = append(handlers, v)
 	}
-	return m
+	return handlers
 }
--- a/pkg/notification/manager.go
+++ b/pkg/notification/manager.go
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2019, Optimizely, Inc. and contributors                        *
+ * Copyright 2019-2020, Optimizely, Inc. and contributors                        *
  *                                                                          *
  * Licensed under the Apache License, Version 2.0 (the "License");          *
  * you may not use this file except in compliance with the License.         *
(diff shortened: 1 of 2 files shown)


pkg/notification/manager.go
```diff
@@ -66,7 +66,17 @@ func (am *AtomicManager) Remove(id int) {
 
 // Send sends the notification to the registered handlers
 func (am *AtomicManager) Send(notification interface{}) {
-	for _, handler := range am.handlers {
+	handlers := am.copyHandlers()
+	for _, handler := range handlers {
```
pawels-optimizely: instead of copying everything at once, can we just do handler := handler inside the for loop ?

msohailhussain: tried didn't work.

pkg/notification/manager.go
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2019, Optimizely, Inc. and contributors                        *
+ * Copyright 2019-2020, Optimizely, Inc. and contributors                        *
```
mikeng13: Haha it's 2020! Still hasn't sunk in yet

pkg/notification/manager.go
```diff
@@ -66,7 +74,20 @@ func (am *AtomicManager) Remove(id int) {
 
 // Send sends the notification to the registered handlers
 func (am *AtomicManager) Send(notification interface{}) {
-	for _, handler := range am.handlers {
+	handlers := am.copyHandlers()
+	for _, handler := range handlers {
+		// copying handler to avoid race condition
```
mikeng13: Move this comment to right before we copy handlers

pkg/notification/manager.go
```diff
@@ -66,7 +74,20 @@ func (am *AtomicManager) Remove(id int) {
...
 func (am *AtomicManager) Send(notification interface{}) {
-	for _, handler := range am.handlers {
+	// copying handler to avoid race condition
+	handlers := am.copyHandlers()
+	for _, handler := range handlers {
 		handler(notification)
 	}
 }
+
+// Copy handlers and return it.
```
mikeng13: nit: Return a copy of the given handlers

pkg/notification/manager.go
```diff
@@ -66,7 +74,20 @@ func (am *AtomicManager) Remove(id int) {
...
+func (am *AtomicManager) copyHandlers() map[uint32]func(interface{}) {
+	am.lock.RLock()
+	defer am.lock.RUnlock()
+	m := make(map[uint32]func(interface{}), len(am.handlers))
+	for k, v := range am.handlers {
+		m[k] = v
+	}
+	return m
+}
```
mikecdavis: We don't really need a map. We can simply copy to a list.

PR #223, opened 2020-01-02
feat: Add observability metrics for event queue

pawels-optimizely:
## Summary
- Added generic metrics through metrics registry, supporting Counter with Add() method and Gauge with Set() method 
- Used these new defined metrics for event dispatcher


pkg/client/factory.go
```diff
@@ -180,6 +190,13 @@ func WithContext(ctx context.Context) OptionFunc {
 	}
 }
 
+// WithMetrics allows user to pass in their own implementation of a metric collector
```
mikeng13: `metrics collector`?

pkg/event/dispatcher.go
```diff
@@ -111,11 +95,11 @@ func (ed *QueueEventDispatcher) flushEvents() {
 
 	retryCount := 0
 
-	ed.metrics.SetQueueSize(ed.eventQueue.Size())
+	ed.stats.Set("queueSize", int64(ed.eventQueue.Size()))
```
mikeng13: Do these keys need to be namespaced by package? Otherwise we'd run into collisions with key names

pawels-optimizely: yes, they can be. Sidedoor has the prefix variable to set the prefix. I can have the same implementation here

pkg/metrics/metrics.go
```diff
@@ -14,49 +14,59 @@
...
 }
 
-// DefaultMetrics stores the actual metrics
-type DefaultMetrics struct {
-	QueueSize         int
-	SuccessFlushCount int64
-	FailFlushCount    int64
-	RetryFlushCount   int64
+// Metrics contains default metrics
+type Metrics struct {
```
mikecdavis: I think the default implementation should be no-op. As it stands now this is not accessible and will only hurt the application performance. Providing the injection interface should be sufficient for consumers to implement. In the future we can provide an out-of-the-box implementation, but it should probably be optional.

pkg/metrics/metrics.go
```diff
@@ -14,49 +14,59 @@
...
+	startTime time.Time
+
+	metricsLock sync.RWMutex
+	metricsData map[string]int64
```
mikecdavis: FWIW, the `expvar.Map` wraps the `sync.Map`.

pkg/metrics/metrics.go
```diff
@@ -14,49 +14,59 @@
...
+import (
+	"sync"
+	"time"
+)
+
+// GenericMetrics provides the interface for the metrics
+type GenericMetrics interface {
+	Inc(key string)
+	Set(key string, val int64)
+	Get(key string) int64
```
mikecdavis: I'm not a fan of this interface because it requires a map lookup which needs to be "thread" safe. This becomes a single point of contention for all metrics being emitted by the SDK. A more explicit interface such as the one we previously [discussed](https://github.com/optimizely/sidedoor/pull/84#pullrequestreview-330226540) would not have to worry about contention for every update, only when first initializing.
```go
type Registry interface {
	NewCounter(name string) Counter
	NewGauge(name string) Gauge
}

type Counter interface {
	Add(delta float64)
}

type Gauge interface {
	Set(delta float64)
}
```
Having the interface _supply_ the "metric" provides a direct reference for the consumer to operate on. The "metric" interface can describe the _type_ of metric such as `Gauge` or `Counter` giving us the flexibility to define more explicit operations on the metric. This "type-safety" gives use the extensibility to add more metrics abstractions such as timers and histograms as the need arises.

A bit of a lesser concern, but relying on strings and having them required for each metric update can also be error prone. When updating the same metric in multiple places, as we do in this PR, it's easy to accidentally use slightly different keys for metrics that should have been the same (e.g. "failFlush" and "failedFlush") 😛

pawels-optimizely: our map is thread safe. We chose to have expvar implementation for sidedoor so we need to keep that implementation in mind. Expvar map implementation is thread safe and it is used already to collect memory. 
 I don't think there is a need to separate counter and gauge interfaces on go-sdk side, a lot of implementations don't make these distinctions , they have exactly same methods for both, and even datadog is using them interchangeably. We have only 3 simple methods: Inc() , Set(), and Get() and I cannot think of anything more useful than these 3 methods. 

Most importantly, we cannot pass registry for expvar variables, we need to pass expvar variables directly already defined, otherwise we will get an error that these are already defined. Expvar variables must be defined on the package level or in a global init(), or using sync.Once . So we cannot define variables on dispatcher, we need to insert these variables to dispatcher or other objects. If we have many metrics, that would mean defining many variables. I think this is the reason why expvar decided to use map to store memory metrics. 

Yeah, you need to be careful with the keys in map, but this is the same concern as with anything else. In your proposal, we can define QueueSize variable and pass "failedFlush" string in a constructor  😀  (too extreme example) . We can also define a new string type and use that instead if we want to prevent that misuse. 

For the sidedoor timers, we could use anything that we wanted because we had a full control. Now we need to pass already defined expvar variables to sidedoor, which may not be that pretty if we use something else than a collection (map). I think map is ok to use for generic go-sdk metrics. 

We can discuss that during our standup or full stack cop.

pkg/client/factory.go
pawels-optimizely: addressed

pkg/metrics/metrics.go
pawels-optimizely: we need to have some implementation regardless. If we don't care to collect metrics on go-sdk yet then I can provide minimal implementation that does nothing (or check every time for nil interface) Also, I don't think injection interface will work really nicely with expvar variables as you need to define them globally - that means in sidedoor and pass them to go-sdk.

pkg/metrics/metrics.go
pawels-optimizely: ok, I have no preferences for locking map or using sync.Map.

pkg/metrics/metrics.go
mikecdavis: > We chose to have expvar implementation for sidedoor so we need to keep that implementation in mind.

expvar is one consumer chosen by sidedoor, but customers will have their own metrics implementations, some using expvar some using others so we shouldn't over index on expvar

> We have only 3 simple methods: Inc() , Set(), and Get() and I cannot think of anything more useful than these 3 methods.

These are methods on a single metric. We're conflating a metric with the collection in this interface.

> Most importantly, we cannot pass registry for expvar variables, we need to pass expvar variables directly already defined, otherwise we will get an error that these are already defined.

This does not change the registry interface. Whether we fetch a "metric" and return it, or fetch a metric and operate on it, we still have to implement under the same constraint.

> I think this is the reason why expvar decided to use map to store memory metrics.

Expvar Memory stats bypass `expvar.Map` and `expvar.Var` entirely. A struct, [runtime.MemStats](https://github.com/golang/go/blob/master/src/runtime/mstats.go#L150) is populated each time the handler is accessed and converted directly to JSON, see [expvar.memstats()](https://github.com/golang/go/blob/master/src/expvar/expvar.go#L357-L361).

> I think map is ok to use for generic go-sdk metrics.

I agree that at the end of the day a map is probably going to exist somewhere, but we don't need to access the map with every update to a metric. Storing a direct reference to the metric takes contention off of the map and allows for a lock-free implementation.

pawels-optimizely: > Most importantly, we cannot pass registry for expvar variables, we need to pass expvar variables directly already defined, otherwise we will get an error that these are already defined.

>>This does not change the registry interface. Whether we fetch a "metric" and return it, or fetch a metric and operate on it, we still have to implement under the same constraint.

It does in my opinion. Unless you have something else in mind. Initializing variables outside of go-sdk won't work. I need to have a collection or a struct that is already initialized and pass that to dispatcher. The generic way of doing that is a map, because we don't know how many metrics we will need in the future.  So, going back to our example, if we define 4 metrics on go-sdk side:
```
type QueueEventDispatcher struct {
	eventQueue     Queue
	eventFlushLock sync.Mutex
	Dispatcher     Dispatcher
	
        queueSize         Metric
	sucessFlush       Metric
	failFlushCounter  Metric
	retryFlushCounter Metric
}
```

then we need to initialize them somewhere else and pass them here. if we have 50 metrics in the future for sidedoor, we would have to do that 50 times , and also probably change the signature of some methods. So fetching and operating is not a problem, the problem is the easiness of initializing and passing them. 

Sure, expvar memory is not using expvar.Map, they are using a global map which also has its own locking for publishing (LoadOrStore() method). No matter how you store for expvar, you will aways lock key by key (at least this is what I see).

pkg/metrics/metrics.go
pawels-optimizely: N/A

pkg/event/processor.go
```diff
@@ -109,6 +112,13 @@ func WithSDKKey(sdkKey string) BPOptionConfig {
 	}
 }
 
+// WithEventDispatcherMetrics sets metrics into the NewProcessor method
+func WithEventDispatcherMetrics(metricsRegistry metrics.Registry) BPOptionConfig {
```
mikecdavis: Nit, this method name should probably be `WithMetricsRegistry(...)`.

pkg/event/dispatcher.go
```diff
@@ -77,7 +72,11 @@ type QueueEventDispatcher struct {
 	eventFlushLock sync.Mutex
 	Dispatcher     Dispatcher
 
-	metrics Metrics
+	// metrics
+	queueSize         metrics.Gauge
+	sucessFlush       metrics.Counter
+	failFlushCounter  metrics.Counter
+	retryFlushCounter metrics.Counter
```
mjc1283: Two of these fields have `Counter` in the field name, but one doesn't - seems weird.

pkg/metrics/metrics.go
```diff
@@ -14,49 +14,51 @@
...
+type Counter interface {
+	Add(delta float64)
 }
 
-// SetQueueSize sets the queue size
-func (m *DefaultMetrics) SetQueueSize(queueSize int) {
-	m.QueueSize = queueSize
+// Gauge interface
+type Gauge interface {
+	Set(delta float64)
```
mjc1283: Is `delta` inappropriate? It's not the change from a prior value right?

PR #224, opened 2020-01-07
refact: Updated polling_manager implementation and unit tests.

yasirfolio3:
## Summary
- Updated SyncConfig to only fetch datafile through http.
- Introduced NewAsyncPollingProjectConfigManager which doesn't initially poll.
- Updated logic to not send notification for initial Datafile.
- Updated polling_manager tests as per the new implementation.


pkg/config/polling_manager.go
```diff
@@ -96,51 +96,46 @@ func WithInitialDatafile(datafile []byte) OptionFunc {
 }
 
 // SyncConfig gets current datafile and updates projectConfig
-func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
+func (cm *PollingProjectConfigManager) SyncConfig() {
```
msohailhussain: downloads datafile

pkg/config/polling_manager.go
```diff
@@ -205,7 +186,40 @@ func NewPollingProjectConfigManager(sdkKey string, pollingMangerOptions ...Optio
...
 	initDatafile := pollingProjectConfigManager.initDatafile
-	pollingProjectConfigManager.SyncConfig(initDatafile) // initial poll
+	if len(initDatafile) == 0 {
+		pollingProjectConfigManager.notificationCenter = registry.GetNotificationCenter(sdkKey)
+		pollingProjectConfigManager.SyncConfig() // initial poll
+	} else {
+		if projectConfig, _ := datafileprojectconfig.NewDatafileProjectConfig(initDatafile); projectConfig != nil {
+			_ = pollingProjectConfigManager.setConfig(projectConfig)
+		}
+		pollingProjectConfigManager.notificationCenter = registry.GetNotificationCenter(sdkKey)
```
msohailhussain: Please add comment, notificationcenter is set later to avoid hard-coded config update notification.

pkg/config/polling_manager.go
```diff
@@ -219,6 +233,32 @@ func (cm *PollingProjectConfigManager) GetConfig() (ProjectConfig, error) {
...
+
+	cm.configLock.Lock()
+	cm.projectConfig = projectConfig
+	if cm.optimizelyConfig != nil {
+		cm.optimizelyConfig = NewOptimizelyConfig(projectConfig)
+	}
+	cm.err = nil
+	cm.configLock.Unlock()
+
+	if cm.notificationCenter != nil {
```
msohailhussain: I think we should separate notificationCenter in a new method.

pkg/config/polling_manager_test.go
```diff
@@ -44,68 +46,168 @@ func newExecGroup() *utils.ExecGroup {
...
+		return err != nil
+	}
+	assertPeriodically(t, evaluationMethod)
+	mockRequester.AssertExpectations(t)
+	eg.TerminateAndWait()
+}
+
 func TestNewPollingProjectConfigManagerWithSimilarDatafileRevisions(t *testing.T) {
 	mockDatafile1 := []byte(`{"revision":"42","botFiltering":true}`)
 	mockDatafile2 := []byte(`{"revision":"42","botFiltering":false}`)
```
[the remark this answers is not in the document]
yasirfolio3: We need the difference in datafiles to make sure the original datafile wasn't changed because of same revision.

pkg/config/polling_manager_test.go
```diff
@@ -44,68 +46,168 @@ func newExecGroup() *utils.ExecGroup {
 	return utils.NewExecGroup(context.Background())
 }
 
+// assertion method to periodically check target function each tick.
+func assertPeriodically(t *testing.T, evaluationMethod func() bool) {
+	assert.Eventually(t, func() bool {
+		return evaluationMethod()
+	}, 500*time.Millisecond, 110*time.Millisecond)
```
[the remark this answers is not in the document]
yasirfolio3: This causes issues in tests where we expect no changes in config after second poll. In such cases, 50ms returns true at 150ms, before the second polling is done which will be at 200ms.

pkg/config/polling_manager_test.go
```diff
@@ -44,68 +46,184 @@ func newExecGroup() *utils.ExecGroup {
...
+		return reflect.DeepEqual(projectConfig1, actual)
+	}
+	assertPeriodically(t, evaluationMethod)
+	mockRequester.AssertExpectations(t)
+	// Check no notification was sent for similar datafile revision
+	assert.Equal(t, uint64(0), atomic.LoadUint64(&numberOfCalls))
+	eg.TerminateAndWait()
+}
+
+func TestNewAsyncPollingProjectConfigManagerWithSimilarDatafileRevisions(t *testing.T) {
```
msohailhussain: what's here you want to test, please add in the comments.

pkg/config/polling_manager_test.go
```diff
@@ -44,68 +46,184 @@ func newExecGroup() *utils.ExecGroup {
...
+	projectConfig1, _ := datafileprojectconfig.NewDatafileProjectConfig(mockDatafile1)
+	mockRequester := new(MockRequester)
+	mockRequester.On("Get", []utils.Header(nil)).Return(mockDatafile2, http.Header{}, http.StatusOK, nil)
+
+	sdkKey := "test_sdk_key"
+	eg := newExecGroup()
+	asyncConfigManager := NewAsyncPollingProjectConfigManager(sdkKey, WithRequester(mockRequester), WithInitialDatafile(mockDatafile1), WithPollingInterval(100*time.Millisecond))
+
+	var numberOfCalls uint64 = 0
+	callback := func(notification notification.ProjectConfigUpdateNotification) {
```
msohailhussain: if we have similar datafile revision, then how it's gonna be called. please explain the reason to declare here.

pkg/config/polling_manager_test.go
```diff
@@ -44,68 +46,184 @@ func newExecGroup() *utils.ExecGroup {
...
+
+	// initialized with hardcoded datafile
+	actual, err := asyncConfigManager.GetConfig()
+	assert.Nil(t, err)
+	assert.NotNil(t, actual)
 	assert.Equal(t, projectConfig1, actual)
+
+	// poll after 100ms
+	eg.Go(asyncConfigManager.Start)
+	evaluationMethod := func() bool {
```
msohailhussain: I think this is not best fit here, just use `sleep`

pkg/config/polling_manager_test.go
```diff
@@ -120,79 +238,130 @@ func TestNewPollingProjectConfigManagerWithLastModifiedDates(t *testing.T) {
...
 	assert.Equal(t, projectConfig1, actual)
 
-	// Sync and check no changes were made to the previous config because of 304 error code
-	configManager.SyncConfig([]byte{})
-	actual, err = configManager.GetConfig()
-	assert.Nil(t, err)
-	assert.NotNil(t, actual)
-	assert.Equal(t, projectConfig1, actual)
+	// poll after 100ms to check no changes were made to the previous config because of 304 error code (first poll)
+	eg.Go(configManager.Start)
```
msohailhussain: Don't use `Start`, use `SyncConfig`

pkg/config/polling_manager_test.go
```diff
@@ -120,79 +238,130 @@ func TestNewPollingProjectConfigManagerWithLastModifiedDates(t *testing.T) {
...
 	actual, err := configManager.GetConfig()
 	assert.Nil(t, err)
 	assert.NotNil(t, actual)
 	assert.Equal(t, projectConfig1, actual)
 
-	configManager.SyncConfig(mockDatafile2)
-	actual, err = configManager.GetConfig()
-	assert.Equal(t, projectConfig2, actual)
+	// poll after 100ms
+	eg.Go(configManager.Start)
```
msohailhussain: Start immediately while you are instantiating object.

pkg/config/polling_manager_test.go
```diff
@@ -120,79 +238,130 @@ func TestNewPollingProjectConfigManagerWithLastModifiedDates(t *testing.T) {
...
-	assert.Equal(t, "42", optimizelyConfig.Revision)
-
-	configManager.SyncConfig(mockDatafile2)
-	optimizelyConfig = configManager.GetOptimizelyConfig()
-	assert.Equal(t, "43", optimizelyConfig.Revision)
+	assert.NotNil(t, actual)
+	assert.Equal(t, projectConfig1, actual)
 
+	// poll after 100ms
+	eg.Go(asyncConfigManager.Start)
```
msohailhussain: don't poll. just use SyncConfig

pkg/config/polling_manager.go
```diff
@@ -157,21 +152,8 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
...
-	if cm.notificationCenter != nil {
-		projectConfigUpdateNotification := notification.ProjectConfigUpdateNotification{
-			Type:     notification.ProjectConfigUpdate,
-			Revision: cm.projectConfig.GetRevision(),
-		}
-		if err = cm.notificationCenter.Send(notification.ProjectConfigUpdate, projectConfigUpdateNotification); err != nil {
-			cmLogger.Warning("Problem with sending notification")
-		}
-	}
+	_ = cm.setConfig(projectConfig)
```
pawels-optimizely: if a function returns an error then we need to check it. It is ok to skip checking in tests, but not in production code.

pkg/config/polling_manager.go
```diff
@@ -205,7 +186,41 @@ func NewPollingProjectConfigManager(sdkKey string, pollingMangerOptions ...Optio
 	}
 
 	initDatafile := pollingProjectConfigManager.initDatafile
-	pollingProjectConfigManager.SyncConfig(initDatafile) // initial poll
+	if len(initDatafile) == 0 {
+		pollingProjectConfigManager.notificationCenter = registry.GetNotificationCenter(sdkKey)
+		pollingProjectConfigManager.SyncConfig() // initial poll
+	} else {
+		if projectConfig, _ := datafileprojectconfig.NewDatafileProjectConfig(initDatafile); projectConfig != nil {
+			_ = pollingProjectConfigManager.setConfig(projectConfig)
```
pawels-optimizely: same comment as above

pkg/config/polling_manager.go
```diff
@@ -205,7 +186,41 @@ func NewPollingProjectConfigManager(sdkKey string, pollingMangerOptions ...Optio
 	}
 
 	initDatafile := pollingProjectConfigManager.initDatafile
-	pollingProjectConfigManager.SyncConfig(initDatafile) // initial poll
+	if len(initDatafile) == 0 {
+		pollingProjectConfigManager.notificationCenter = registry.GetNotificationCenter(sdkKey)
```
pawels-optimizely: we initialize notification center in "else" so I think it can be done before if statement

pkg/config/polling_manager.go
```diff
@@ -205,7 +186,41 @@ func NewPollingProjectConfigManager(sdkKey string, pollingMangerOptions ...Optio
...
+		pollingProjectConfigManager.SyncConfig() // initial poll
+	} else {
+		if projectConfig, _ := datafileprojectconfig.NewDatafileProjectConfig(initDatafile); projectConfig != nil {
```
pawels-optimizely: lines 193 -198 are repeated in NewAsync, maybe it will be better to create a small help function for that.

pkg/config/polling_manager.go
yasirfolio3: both initialisations serve different purposes, the one in `if` condition is before the `SyncConfig` call which makes sure notification is sent, the one in `else` is after the config has been set in which case notification won't be sent.

pkg/config/polling_manager.go
pawels-optimizely: I would remove locking mechanism here and minimize the number of calls to locking. 
of course cm.sendConfigUpdateNotification() will need to be called in  SyncConfig then

pkg/config/polling_manager.go
pawels-optimizely: cm.setConfig function should not send notification

pkg/config/polling_manager.go
```diff
@@ -219,6 +234,35 @@ func (cm *PollingProjectConfigManager) GetConfig() (ProjectConfig, error) {
 	return cm.projectConfig, nil
 }
 
+func (cm *PollingProjectConfigManager) setConfig(projectConfig ProjectConfig) error {
```
pawels-optimizely: I would remove locking mechanism here and minimize the number of calls to locking.
of course cm.sendConfigUpdateNotification() will need to be called in SyncConfig then

pkg/config/polling_manager.go
msohailhussain: then where notification should be sent? I think setConfig has direct dependency of notification that's why it's sending notification in there.

pawels-optimizely: notification can be sent in SysConfig() when setConfig returns nil error

pkg/config/polling_manager.go
```diff
@@ -256,3 +261,35 @@ func (cm *PollingProjectConfigManager) RemoveOnProjectConfigUpdate(id int) error
...
+
+func (cm *PollingProjectConfigManager) setInitialDatafile(datafile []byte) {
+	if len(datafile) != 0 {
+		if projectConfig, err := datafileprojectconfig.NewDatafileProjectConfig(datafile); projectConfig != nil {
+			cm.configLock.Lock()
+			defer cm.configLock.Unlock()
+			cm.setConfig(projectConfig)
+			cm.err = nil
+		} else {
+			cm.err = err
```
pawels-optimizely: I think , to be safe, we need to lock cm.err too.

pkg/config/polling_manager.go
```diff
@@ -157,21 +152,9 @@ func (cm *PollingProjectConfigManager) SyncConfig(datafile []byte) {
...
-	if cm.notificationCenter != nil {
-		projectConfigUpdateNotification := notification.ProjectConfigUpdateNotification{
-			Type:     notification.ProjectConfigUpdate,
-			Revision: cm.projectConfig.GetRevision(),
-		}
-		if err = cm.notificationCenter.Send(notification.ProjectConfigUpdate, projectConfigUpdateNotification); err != nil {
-			cmLogger.Warning("Problem with sending notification")
-		}
-	}
+	cm.sendConfigUpdateNotification()
```
pawels-optimizely: I think cm.setConfig() should still return an error, and we need to sendConfigUpdateNotification only when error is not nil

pkg/config/polling_manager.go
```diff
@@ -151,10 +151,12 @@ func (cm *PollingProjectConfigManager) SyncConfig() {
...
-	cmLogger.Debug(fmt.Sprintf("New datafile set with revision: %s. Old revision: %s", projectConfig.GetRevision(), previousRevision))
-	cm.setConfig(projectConfig)
-	closeMutex(nil)
-	cm.sendConfigUpdateNotification()
+	err = cm.setConfig(projectConfig)
+	if err == nil {
+		cmLogger.Debug(fmt.Sprintf("New datafile set with revision: %s. Old revision: %s", projectConfig.GetRevision(), previousRevision))
+		cm.sendConfigUpdateNotification()
+	}
+	closeMutex(err)
```
pawels-optimizely: closing mutex needs to happen before sending notification.

PR #232, opened 2020-01-17
feat: set specific error from default project config manager when CDN returns 403 response

pawels-optimizely:
## Summary

Assigning error to a predefined error variable when CDN returns 403 response helps consumers such as sidedoor. Currently these consumers have to search the error message string for the text "403", which is not really convenient.

### DIFF
--- a/pkg/config/polling_manager.go
+++ b/pkg/config/polling_manager.go
@@ -46,7 +46,7 @@ const LastModified = "Last-Modified"
 const DatafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json"
 
 // Err403Forbidden is 403Forbidden specific error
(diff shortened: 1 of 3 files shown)


pkg/config/polling_manager.go
```diff
@@ -45,6 +45,9 @@ const LastModified = "Last-Modified"
 // DatafileURLTemplate is used to construct the endpoint for retrieving the datafile from the CDN
 const DatafileURLTemplate = "https://cdn.optimizely.com/datafiles/%s.json"
 
+// Err403Forbidden is 403Forbidden specific error
+var Err403Forbidden = errors.New("unable to fetch fresh datafile, reason (http status code): 403 Forbidden")
```
aliabbasrizvi: It will be good to mention that people confirm their SDK key again.

PR #234, opened 2020-01-23
feat: add valid datafile version check

pawels-optimizely:
## Summary
to be compatible with other sdk's, we are adding a check for a valid version of datafile.  go-sdk supports only version 4 of datafile, same as the family of more recent sdk's.

### DIFF
--- a/pkg/config/datafileprojectconfig/config.go
+++ b/pkg/config/datafileprojectconfig/config.go
@@ -185,13 +185,13 @@ func (c DatafileProjectConfig) GetGroupByID(groupID string) (entities.Group, err
 func NewDatafileProjectConfig(jsonDatafile []byte) (*DatafileProjectConfig, error) {
 	datafile, err := Parse(jsonDatafile)
 	if err != nil {
-		logger.Error("error parsing datafile", err)
+		logger.Error("Error parsing datafile", err)
 		return nil, err
 	}
 
 	if !inSlice(datafileVersions, datafile.Version) {
 		err = errors.New("unsupported datafile version")
-		logger.Error(fmt.Sprintf("version %s of datafile not supported", datafile.Version), err)
+		logger.Error(fmt.Sprintf("Version %s of datafile not supported", datafile.Version), err)
 		return nil, err
 	}
 
--- a/pkg/config/datafileprojectconfig/config.go
+++ b/pkg/config/datafileprojectconfig/config.go
@@ -28,15 +28,8 @@ import (
 
 var logger = logging.GetLogger("DatafileProjectConfig")
 
-var datafileVersions = []string{"4"}
-
-func inSlice(a []string, x string) bool {
-	for _, n := range a {
-		if x == n {
-			return true
-		}
-	}
-	return false
+var datafileVersions = map[string]struct{}{
+	"4": {},
 }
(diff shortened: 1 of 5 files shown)


pkg/config/datafileprojectconfig/config.go
```diff
@@ -27,6 +28,17 @@ import (
 
 var logger = logging.GetLogger("DatafileProjectConfig")
 
+var datafileVersions = []string{"4"}
```
mikeng13: Wouldn’t a map make this simpler since we wouldn’t have to iterate?

pkg/config/datafileprojectconfig/config.go
```diff
@@ -173,7 +185,13 @@ func (c DatafileProjectConfig) GetGroupByID(groupID string) (entities.Group, err
...
 	datafile, err := Parse(jsonDatafile)
 	if err != nil {
-		logger.Error("Error parsing datafile.", err)
+		logger.Error("Error parsing datafile", err)
+		return nil, err
+	}
+
+	if !inSlice(datafileVersions, datafile.Version) {
+		err = errors.New("unsupported datafile version")
+		logger.Error(fmt.Sprintf("Version %s of datafile not supported", datafile.Version), err)
```
aliabbasrizvi: Will be useful to say unable to initialize/update or something along those lines.

pkg/config/datafileprojectconfig/config.go
mikecdavis: Consider using a `map[string]struct{}`, then you can check via `_, ok := datafileVersions[datafile.Version]` and we don't need the `inSlice` method defined either.

https://stackoverflow.com/questions/10485743/contains-method-for-a-slice

pkg/config/datafileprojectconfig/config_test.go
```diff
@@ -30,11 +30,16 @@ func TestNewDatafileProjectConfigNil(t *testing.T) {
 	projectConfig, err := NewDatafileProjectConfig(nil)
 	assert.NotNil(t, err)
 	assert.Nil(t, projectConfig)
+
+	jsonDatafileStr := `{"accountID": "123", "revision": "1", "projectId": "12345", "version": "2"}`
+	projectConfig, err = NewDatafileProjectConfig([]byte(jsonDatafileStr))
+	assert.NotNil(t, err)
```
mikecdavis: Nit, I think we should be standardizing on `assert.Error(t, err)`

pkg/config/static_manager_test.go
```diff
@@ -39,7 +39,11 @@ func TestNewStaticProjectConfigManagerFromPayload(t *testing.T) {
 	configManager, err := NewStaticProjectConfigManagerFromPayload(mockDatafile)
 	assert.NotNil(t, err)
 
-	mockDatafile = []byte(`{"accountId":"42","projectId":"123"}`)
+	mockDatafile = []byte(`{"accountId":"42","projectId":"123",}`)
+	configManager, err = NewStaticProjectConfigManagerFromPayload(mockDatafile)
+	assert.NotNil(t, err)
```
mikecdavis: same here with `assert.Error(t, err)`

PR #235, opened 2020-01-24
Return variables as an interface{} from GetAllFeatureVariables.

mikecdavis:
## Summary
* Return typed variables to be more consistent with the discrete methods.

This change is also going to be consumed by Agent to be able to provide typed responses.


mikeng13:
@pawels-optimizely other SDKs don't offer this API yet. 

This could be construed as a breaking change though since it is changing the signature. Though I don't feel too strongly about it, just wanna make sure we're ok with that.

PR #236, opened 2020-01-27
feat: Add blocking timeout in polling config manager.

yasirfolio3:
## Summary
This introduces blocking timeout in PollingConfigManager. Config Manager will block any calling go-routine until the config is available or the timeout occurs.

### DIFF
--- a/pkg/config/polling_manager_test.go
+++ b/pkg/config/polling_manager_test.go
@@ -163,6 +163,7 @@ func TestGetConfigForPollingConfigManagerWaitsForTimeoutForInvalidDatafile(t *te
 
 	start := time.Now()
 	expectedExecutionTime := start.Add(blockingTimeout)
+	// blocking api for specified time.
 	_, _ = configManager.GetConfig()
 	actualExecutionTime := start.Add(time.Since(start))
 	assert.WithinDuration(t, expectedExecutionTime, actualExecutionTime, 10*time.Millisecond)
@@ -181,6 +182,7 @@ func TestGetConfigForAsyncPollingConfigManagerWaitsForTimeoutForInvalidDatafile(
 	configManager.SyncConfig()
 	start := time.Now()
 	expectedExecutionTime := start.Add(blockingTimeout)
+	// blocking api for specified time.
 	_, _ = configManager.GetConfig()
 	actualExecutionTime := start.Add(time.Since(start))
 	assert.WithinDuration(t, expectedExecutionTime, actualExecutionTime, 10*time.Millisecond)
--- a/pkg/config/polling_manager_test.go
+++ b/pkg/config/polling_manager_test.go
@@ -119,12 +119,13 @@ func TestSyncConfigFetchesDatafileUsingRequester(t *testing.T) {
 }
 
 func TestNewPollingProjectConfigManagerWithNull(t *testing.T) {
+	blockingTimeout := 5 * time.Millisecond
 	mockDatafile := []byte("NOT-VALID")
 	mockRequester := new(MockRequester)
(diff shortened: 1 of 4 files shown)


msohailhussain:
converage is decreased, @yasirfolio3 let's discuss how coverage can be increased.

pkg/utils/channel_utils.go
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package utils //
+package utils
+
+import "time"
+
+// IsChannelClosed will check if the given channel is closed
+func IsChannelClosed(ch chan struct{}) bool {
```
mjc1283: Nit: can we use read-only channel type `<-chan struct{}`? And also below in `WaitForChannelToCloseOrTimeout`?

pkg/config/polling_manager_test.go
```diff
@@ -146,6 +150,127 @@ func TestNewAsyncPollingProjectConfigManagerWithNullDatafile(t *testing.T) {
...
+
+	configManager := NewPollingProjectConfigManager(sdkKey, WithRequester(mockRequester), WithBlockingTimeout(blockingTimeout))
+	mockRequester.AssertExpectations(t)
+
+	start := time.Now()
+	expectedExecutionTime := start.Add(blockingTimeout)
+	_, _ = configManager.GetConfig()
+	actualExecutionTime := start.Add(time.Since(start))
+	assert.WithinDuration(t, expectedExecutionTime, actualExecutionTime, 10*time.Millisecond)
+}
```
mjc1283: This test case and the others with 50ms blocking timeout seem to have the potential to be flaky, and will run very slowly. I will think about other ways to test this behavior - @pawels-optimizely @mikeng13 @mikecdavis any ideas?

yasirfolio3:
> Nit in PR description:
> 
> > Config Manager will block the main thread until the config is available or the timeout occurs
> 
> What is the "main thread"? It will block any calling goroutine, right?

correct, thanks for the correction.

pkg/config/polling_manager_test.go
```diff
@@ -146,6 +150,129 @@ func TestNewAsyncPollingProjectConfigManagerWithNullDatafile(t *testing.T) {
...
+
+	configManager := NewPollingProjectConfigManager(sdkKey, WithRequester(mockRequester), WithBlockingTimeout(blockingTimeout))
+	mockRequester.AssertExpectations(t)
+
+	start := time.Now()
+	expectedExecutionTime := start.Add(blockingTimeout)
+	// blocking api for specified time.
+	_, _ = configManager.GetConfig()
+	actualExecutionTime := start.Add(time.Since(start))
+	assert.WithinDuration(t, expectedExecutionTime, actualExecutionTime, 10*time.Millisecond)
```
mjc1283: @yasirfolio3 @msohailhussain Following up on my earlier comment.

To limit flakiness I suggest we remove this `assert.WithinDuration`. Instead, we can implement a timeout on the total execution time of this test case, failing when it exceeds a pretty large duration, like 1-2 seconds. This lets us be resilient to varying execution time of `GetConfig`, while still ensuring that our test fails when the implementation of the timeout inside `GetConfig` is broken.

pkg/config/polling_manager_test.go
```diff
@@ -146,6 +150,129 @@ func TestNewAsyncPollingProjectConfigManagerWithNullDatafile(t *testing.T) {
 	mockRequester.AssertExpectations(t)
 }
 
+func TestGetConfigForPollingConfigManagerWaitsForTimeoutForInvalidDatafile(t *testing.T) {
+	blockingTimeout := 50 * time.Millisecond
```
mjc1283: I suggest reducing this to 1 millisecond, combined with my other suggestion, there is no need for this to be so high.

PR #238, opened 2020-02-07
ci(travis): add mdspell

juancarlostong:


.travis.yml
```diff
@@ -14,15 +14,26 @@ stages:
...
         email: false
 
+    - stage: 'Lint markdown files'
+      os: linux
+      language: generic
+      before_install: skip
+      install:
+        - npm i -g markdown-spellcheck
+      before_script:
+        - wget --quiet https://raw.githubusercontent.com/optimizely/mdspell-config/master/.spelling
```
pawels-optimizely: I thought we wanted to keep that file locally. Are we going to make a change to this global file all the time?

juancarlostong: we want to use global so we dont have to maintain redundant entries for all sdk repos.

eg.

optimizely is special dictionary

should be for every sdk.. lets just put it in one place

PR #241, opened 2020-03-12
feature (logging): intial checkin for adding log producer per sdk key

thomaszurkan-optimizely:


thomaszurkan-optimizely:
> I started to review this but I saw your comment that this is not done yet.
> I have a few suggestions:
> 
> 1. I would separate context passing with sdk key just for logger from required sdk key needed by a function. Although in our case it's a bit redundant, but later if we decide to cache our logger by project id then  we won't have to change that much  of a code.  So, polling manager would take sdk key - because it needs for polling, and whatever context it needs to take for logging.
> 2. Please update main.go in examples.

This is good feedback.  What I was thinking was that we have the "activation context."  That could hold onto anything needed for activation sdk key, project id, etc.  Furthermore, you will see that composite objects take a context and smaller granularity objects just take a logger. I don't have a problem with the redundant parameters.  I'm just telling you what I was thinking.  Anyone else have an opinion on this?

pkg/event/processor.go
```diff
@@ -66,7 +66,9 @@ const DefaultEventFlushInterval = 30 * time.Second
 
 const maxFlushWorkers = 1
 
-var pLogger = logging.GetLogger("EventProcessor")
+func pLogger(c context.Context) logging.OptimizelyLogProducer {
+	return logging.GetLogger(c, "EventProcessor")
+}
```
mikecdavis: Similar comment here, logger should be a field on the `BatchEventProcessor`

pkg/logging/logger.go
```diff
@@ -18,21 +18,47 @@
...
+
+func getLogger(sdkKey string) OptimizelyLogConsumer {
+	if val, ok := loggersCache.Load(sdkKey);ok && val != nil {
+		if lp, ok := val.(OptimizelyLogConsumer); ok {
+			return lp
+		}
+	}
+	return nil
+}
+
```
mikecdavis: I don't think this is necessary and would constitute a breaking change.

pkg/logging/logger.go
```diff
@@ -47,35 +73,31 @@ const (
...
+		}
+	}
 }
 
 // GetLogger returns a log producer with the given name
-func GetLogger(name string) OptimizelyLogProducer {
+func GetLogger(ctx context.Context, name string) OptimizelyLogProducer {
 	return NamedLogProducer{
+		ctx: ctx,
 		fields: map[string]interface{}{"name": name},
```
mikecdavis: The main crux of this PR is to make the SDK available to the OptimizelyLogConsumer. The `fields` member of the `NamedLogProducer` is meant to capture this type of arbitrary KV pairs to "enrich" the logs.

See https://github.com/optimizely/agent/blob/master/pkg/optimizely/logger.go#L54-L56 for usage of fields in an implementation of a OptimizelyLogConsumer.

pkg/logging/logger.go
```diff
@@ -102,8 +124,17 @@ func (p NamedLogProducer) Error(message string, err interface{}) {
...
-	mutex.Unlock()
+	if logger := getLogger(GetSdkKey(p.ctx)); logger != nil {
+		loggerLock.Lock()
+		logger.Log(logLevel, message, p.fields)
+		loggerLock.Unlock()
+	} else {
+		loggerLock.Lock()
+		defaultLogConsumer.Log(logLevel, message, p.fields)
+		loggerLock.Unlock()
+	}
```
mikecdavis: Per the pervious comment about using `fields` to propagate the SDK key, I don't think we need to support multiple SDK consumers anymore.

pkg/client/factory.go
```diff
@@ -73,29 +74,29 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 		ctx = context.Background()
 	}
 
+	ctx = context.WithValue(ctx, logging.SdkKey, f.SDKKey)
+
 	eg := utils.NewExecGroup(ctx)
-	appClient := &OptimizelyClient{execGroup: eg, notificationCenter: registry.GetNotificationCenter(f.SDKKey)}
+	appClient := &OptimizelyClient{execGroup: eg, notificationCenter: registry.GetNotificationCenter(f.SDKKey), logger:logging.GetLogger(ctx, "OptimizelyClient")}
```
mikecdavis: Nit, this line is getting pretty long, can we break out into multiple lines?

pkg/event/processor.go
mikecdavis: Looks like I lost my original comment somewhere, so to clarify I think we should be consistent and store the logger on the structs and not look them up dynamically with each log statement. This adds unnecessary overhead to each log statement.

PR #242, opened 2020-03-19
feature (logging): Logging sdk key with every log message.

thomaszurkan-optimizely:
The sdk key is passed to composite classes and classes that create other objects that require logging.  Small grained classes just accept a logger.  

The logger logs sdk key mapped value by default along with name.
There are setters and getters for the sdk key log mapping.
Added unit test for sdk key in log.

Also updated the travis unit tests from 1.8 to 1.10. and from 1.13 to 1.14.

### DIFF
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -20,12 +20,12 @@ package client
 import (
 	"context"
 	"errors"
-	"github.com/optimizely/go-sdk/pkg/logging"
 	"time"
 
 	"github.com/optimizely/go-sdk/pkg/config"
 	"github.com/optimizely/go-sdk/pkg/decision"
 	"github.com/optimizely/go-sdk/pkg/event"
+	"github.com/optimizely/go-sdk/pkg/logging"
 	"github.com/optimizely/go-sdk/pkg/metrics"
 	"github.com/optimizely/go-sdk/pkg/registry"
 	"github.com/optimizely/go-sdk/pkg/utils"
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -20,6 +20,7 @@ package client
 import (
 	"context"
 	"errors"
+	"github.com/optimizely/go-sdk/pkg/logging"
 	"time"
 
 	"github.com/optimizely/go-sdk/pkg/config"
@@ -73,8 +74,10 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 		ctx = context.Background()
 	}
 
-	eg := utils.NewExecGroup(ctx)
-	appClient := &OptimizelyClient{execGroup: eg, notificationCenter: registry.GetNotificationCenter(f.SDKKey)}
+	eg := utils.NewExecGroup(logging.GetLogger(f.SDKKey, "ExecGroup"), ctx)
+	appClient := &OptimizelyClient{execGroup: eg,
+		notificationCenter: registry.GetNotificationCenter(f.SDKKey),
+		logger: logging.GetLogger(f.SDKKey, "OptimizelyClient")}
 
 	if f.configManager != nil {
 		appClient.ConfigManager = f.configManager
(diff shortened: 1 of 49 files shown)


pkg/event/processor.go
```diff
@@ -129,6 +127,8 @@ func NewBatchEventProcessor(options ...BPOptionConfig) *BatchEventProcessor {
 		opt(p)
 	}
 
+	p.logger = logging.GetLogger(p.sdkKey, "BatchEventProcessor")
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: no, because the sdk key is part of the config options.  So, first I run my config options, they, I grab the sdk key, then I get the logger with the sdk key.

pkg/client/factory.go
```diff
@@ -73,8 +74,10 @@ func (f OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClien
 		ctx = context.Background()
 	}
 
-	eg := utils.NewExecGroup(ctx)
-	appClient := &OptimizelyClient{execGroup: eg, notificationCenter: registry.GetNotificationCenter(f.SDKKey)}
+	eg := utils.NewExecGroup(ctx, logging.GetLogger(f.SDKKey, "ExecGroup"))
+	appClient := &OptimizelyClient{execGroup: eg,
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: ran it.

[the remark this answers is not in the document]
thomaszurkan-optimizely: so, by default you will have a sdk key mapping of a log mapping which goes to optimizely-1,2,3, etc.  To override, you would simply call UseSdkKeyForLogging(sdkKey) before creating your client factory.

pkg/logging/level_log_consumer.go
```diff
@@ -33,8 +33,21 @@ type FilteredLevelLogConsumer struct {
...
+		for _, v := range fields {
+			if s, ok := v.(string); ok {
+				_, _ = messBuilder.WriteString("[")
+				_, _ = messBuilder.WriteString(s)
+				_, _ = messBuilder.WriteString("]")
+			}
+		}
+		_, _ = messBuilder.WriteString(" ")
+		_, _ = messBuilder.WriteString(message)
+		l.logger.Println(messBuilder.String())
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: now using optimizely-1,2,3,4,5,etc.  as a mapping.

pkg/logging/level_log_consumer.go
```diff
@@ -33,8 +33,21 @@ type FilteredLevelLogConsumer struct {
 func (l *FilteredLevelLogConsumer) Log(level LogLevel, message string, fields map[string]interface{}) {
 	if l.level <= level {
 		// prepends the name and log level to the message
-		message = fmt.Sprintf("[%s][%s] %s", level, fields["name"], message)
-		l.logger.Println(message)
+		messBuilder := strings.Builder{}
+		_, _ = messBuilder.WriteString("[")
```
pawels-optimizely: "_, _" is not needed here, and errors are not checked, I think fmt.Sprintf will be cleaner here.

pkg/decision/bucketer/murmurhashbucketer.go
```diff
@@ -42,20 +42,22 @@ type Bucketer interface {
 // MurmurhashBucketer generates the bucketing value using the mmh3 algorightm
 type MurmurhashBucketer struct {
 	hashSeed uint32
+	logger   logging.OptimizelyLogProducer
 }
 
 // NewMurmurhashBucketer returns a new instance of the murmurhash bucketer
-func NewMurmurhashBucketer(hashSeed uint32) *MurmurhashBucketer {
+func NewMurmurhashBucketer(sdkKey string, hashSeed uint32) *MurmurhashBucketer {
```
pawels-optimizely: little weird that NewMurmurhashBucketer takes sdkKey .

pkg/logging/level_log_consumer.go
thomaszurkan-optimizely: at the consumer level, there is no telling how many interfaces have been added.  so, a sprintf is not appropriate here.

pkg/decision/bucketer/murmurhashbucketer.go
thomaszurkan-optimizely: good catch.  at that level, it should just take a logger. fixed.

pkg/logging/level_log_consumer.go
thomaszurkan-optimizely: also, the _,_ was a result of go lint.  our linter required it.

pawels-optimizely: I think that linters complain about not checking an error. Anyway, WriteString is returning nil error all the time, so there is really no point of checking that error. 

I wanted to simplify this:
```
                 _, _ = messBuilder.WriteString("[")
		_, _ = messBuilder.WriteString(level.String())
		_, _ = messBuilder.WriteString("]")
```
with
``` 
message = fmt.Sprintf("[%d]", level )
_, _ = messBuilder.WriteString(message)
```
maybe it's more readable

PR #247, opened 2020-03-26
feature (multi-rollout): Added support for multiple rollouts.

yasirfolio3:
## Summary
Added support for multiple rollouts in `rollout_service`, this includes:
- Evaluating next available experiment if audience evaluation fails for previous.
- Falling back to the last experiment evaluation if bucketing fails for any of the experiments.

### DIFF
--- a/pkg/decision/rollout_service.go
+++ b/pkg/decision/rollout_service.go
@@ -44,45 +44,24 @@ func NewRolloutService(sdkKey string) *RolloutService {
 
 // GetDecision returns a decision for the given feature and user context
 func (r RolloutService) GetDecision(decisionContext FeatureDecisionContext, userContext entities.UserContext) (FeatureDecision, error) {
+
 	featureDecision := FeatureDecision{
 		Source: Rollout,
 	}
 	feature := decisionContext.Feature
 	rollout := feature.Rollout
-	if rollout.ID == "" {
-		featureDecision.Reason = reasons.NoRolloutForFeature
-		return featureDecision, nil
+
+	evaluateConditionTree := func(experiment entities.Experiment) bool {
+		condTreeParams := entities.NewTreeParameters(&userContext, decisionContext.ProjectConfig.GetAudienceMap())
+		evalResult, _ := r.audienceTreeEvaluator.Evaluate(experiment.AudienceConditionTree, condTreeParams)
+		if !evalResult {
+			featureDecision.Reason = reasons.FailedRolloutTargeting
+			r.logger.Debug(fmt.Sprintf(`User "%s" failed targeting for feature rollout with key "%s".`, userContext.ID, feature.Key))
+		}
+		return evalResult
 	}
 
-	numberOfExperiments := len(rollout.Experiments)
-	if numberOfExperiments == 0 {
-		featureDecision.Reason = reasons.RolloutHasNoExperiments
-		return featureDecision, nil
-	}
-
-	index := 0
-	for index < numberOfExperiments {
-		experiment := rollout.Experiments[index]
-		experimentDecisionContext := ExperimentDecisionContext{
-			Experiment:    &experiment,
-			ProjectConfig: decisionContext.ProjectConfig,
-		}
-		if experiment.AudienceConditionTree != nil {
-			condTreeParams := entities.NewTreeParameters(&userContext, decisionContext.ProjectConfig.GetAudienceMap())
-			evalResult, _ := r.audienceTreeEvaluator.Evaluate(experiment.AudienceConditionTree, condTreeParams)
-			if !evalResult { // Evaluate this user for the next rule
-				featureDecision.Reason = reasons.FailedRolloutTargeting
-				r.logger.Debug(fmt.Sprintf(`User "%s" failed targeting for feature rollout with key "%s".`, userContext.ID, feature.Key))
-				index++
-				continue
-			}
-		}
-		decision, _ := r.experimentBucketerService.GetDecision(experimentDecisionContext, userContext)
-		if decision.Variation == nil && index < numberOfExperiments-1 {
-			// Evaluate fall back rule / last rule now
-			index = numberOfExperiments - 1
-			continue
-		}
+	getFeatureDecision := func(experiment entities.Experiment, decision ExperimentDecision) (FeatureDecision, error) {
 		// translate the experiment reason into a more rollouts-appropriate reason
 		switch decision.Reason {
 		case reasons.NotBucketedIntoVariation:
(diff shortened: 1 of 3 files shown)


pkg/decision/rollout_service.go
```diff
@@ -61,38 +61,43 @@ func (r RolloutService) GetDecision(decisionContext FeatureDecisionContext, user
...
-	// if user fails rollout targeting rule we return out of it
-	if experiment.AudienceConditionTree != nil {
-		condTreeParams := entities.NewTreeParameters(&userContext, decisionContext.ProjectConfig.GetAudienceMap())
-		evalResult, _ := r.audienceTreeEvaluator.Evaluate(experiment.AudienceConditionTree, condTreeParams)
-		if !evalResult {
-			featureDecision.Reason = reasons.FailedRolloutTargeting
-			rsLogger.Debug(fmt.Sprintf(`User "%s" failed targeting for feature rollout with key "%s".`, userContext.ID, feature.Key))
-			return featureDecision, nil
+	index := 0
+	for index < numberOfExperiments {
```
msohailhussain: it should be numberOfExperiments-1

pkg/decision/rollout_service.go
```diff
@@ -61,38 +61,43 @@ func (r RolloutService) GetDecision(decisionContext FeatureDecisionContext, user
...
+				rsLogger.Debug(fmt.Sprintf(`User "%s" failed targeting for feature rollout with key "%s".`, userContext.ID, feature.Key))
+				index++
+				continue
+			}
+		}
+		decision, _ := r.experimentBucketerService.GetDecision(experimentDecisionContext, userContext)
+		if decision.Variation == nil && index < numberOfExperiments-1 {
+			// Evaluate fall back rule / last rule now
+			index = numberOfExperiments - 1
+			continue
```
msohailhussain: it should be `break`

msohailhussain: Please add `everyoneElse` rule.

pkg/decision/rollout_service.go
```diff
@@ -22,21 +22,22 @@ import (
 
 	"github.com/optimizely/go-sdk/pkg/decision/evaluator"
 	"github.com/optimizely/go-sdk/pkg/decision/reasons"
-	"github.com/optimizely/go-sdk/pkg/entities"
 	"github.com/optimizely/go-sdk/pkg/logging"
+
+	"github.com/optimizely/go-sdk/pkg/entities"
```
msohailhussain: please revert it.

pkg/decision/rollout_service.go
```diff
@@ -60,38 +61,43 @@ func (r RolloutService) GetDecision(decisionContext FeatureDecisionContext, user
...
+		experimentDecisionContext := ExperimentDecisionContext{
+			Experiment:    &experiment,
+			ProjectConfig: decisionContext.ProjectConfig,
+		}
+		if experiment.AudienceConditionTree != nil {
+			condTreeParams := entities.NewTreeParameters(&userContext, decisionContext.ProjectConfig.GetAudienceMap())
+			evalResult, _ := r.audienceTreeEvaluator.Evaluate(experiment.AudienceConditionTree, condTreeParams)
+			if !evalResult { // Evaluate this user for the next rule
+				featureDecision.Reason = reasons.FailedRolloutTargeting
+				r.logger.Debug(fmt.Sprintf(`User "%s" failed targeting for feature rollout with key "%s".`, userContext.ID, feature.Key))
```
msohailhussain: are we loggin in other sdks?

pkg/decision/rollout_service_test.go
```diff
@@ -26,25 +26,24 @@ import (
...
-	mockAudienceTreeEvaluator     *MockAudienceTreeEvaluator
-	mockExperimentService         *MockExperimentDecisionService
-	testExperimentDecisionContext ExperimentDecisionContext
-	testFeatureDecisionContext    FeatureDecisionContext
-	testConditionTreeParams       *entities.TreeParameters
-	testUserContext               entities.UserContext
+	mockConfig                        *mockProjectConfig
+	mockAudienceTreeEvaluator         *MockAudienceTreeEvaluator
+	mockExperimentService             *MockExperimentDecisionService
+	testExperiment1112DecisionContext ExperimentDecisionContext
```
msohailhussain: what is it 1112?

pkg/decision/rollout_service_test.go
```diff
@@ -89,27 +124,98 @@ func (s *RolloutServiceTestSuite) TestGetDecisionHappyPath() {
...
 	}
 	expectedFeatureDecision := FeatureDecision{
+		Experiment: testExp1118,
+		Variation:  &testExp1118Var2224,
+		Source:     Rollout,
+		Decision:   Decision{Reason: reasons.BucketedIntoRollout},
+	}
+	decision, _ := testRolloutService.GetDecision(s.testFeatureDecisionContext, s.testUserContext)
+	s.Equal(expectedFeatureDecision, decision)
+	s.mockAudienceTreeEvaluator.AssertExpectations(s.T())
```
msohailhussain: Can we also assert, number of rolloutrules `experiments` were in it? and the returned mocked roleout rule was the last one. Also throw `exception` for all other roleout rules so I can be sure, it didn't iterate once fallen into audience evaluation but didn't bucket.

pkg/decision/rollout_service_test.go
yasirfolio3: it represents the `experimentID`.

pkg/decision/rollout_service_test.go
yasirfolio3: Since there are 3 rollout rules `experiments` in this test, we have only mocked response for `Evaluate` and `GetDecision` methods of 1st and last rule, This makes sure that if these methods were called for any other experiments, the test will fail. We assert that the returned rollout rule was the last one already by providing `testExp1118 ` in expectedFeatureDecision.

pkg/decision/rollout_service.go
yasirfolio3: yes we do.

pkg/decision/rollout_service.go
```diff
@@ -30,25 +30,61 @@ import (
...
 // GetDecision returns a decision for the given feature and user context
 func (r RolloutService) GetDecision(decisionContext FeatureDecisionContext, userContext entities.UserContext) (FeatureDecision, error) {
+
 	featureDecision := FeatureDecision{
 		Source: Rollout,
 	}
 	feature := decisionContext.Feature
 	rollout := feature.Rollout
+
+	evaluateConditionTree := func(experiment entities.Experiment) bool {
```
pawels-optimizely, line 54: since this is a help function, can you try passing by reference instead of value?

PR #250, opened 2020-04-01
feat: implementing OptimizelyJSON

pawels-optimizely:
# Summary
- implementing OptimizelyJSON object and its API's needed to support json access for new public SDK API's


pkg/optimizelyjson/optimizely_json.go
```diff
@@ -0,0 +1,98 @@
...
+	jsonBytes, err := json.Marshal(optlyJson.jsonRepr)
+	if err != nil {
+		return "", err
+	}
+
+	return string(jsonBytes), nil
+}
+
+// ToDict returns the native representation of json (map of interface)
+func (optlyJson OptimizelyJSON) ToDict() map[string]interface{} {
```
msohailhussain, line 48: Doesn't it sound confusing to call it `ToDict` while go uses `map` for `key/value` pairs.

pkg/optimizelyjson/optimizely_json.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// OptimizelyJSON holds the underlying structure of the object
+type OptimizelyJSON struct {
+	jsonRepr map[string]interface{}
```
msohailhussain, line 29: `jsonRepr` looks like holding the formatted value but I believe, it should be independent of any `Representation`, i mean the variable name should be something called `data` it's just a suggestion.

pkg/optimizelyjson/optimizely_json.go
pawels-optimizely: > Doesn't it sound confusing to call it `ToDict` while go uses `map` for `key/value` pairs.

yeah, I wanted to come up with something generic to all strictly typed languages. But maybe choosing something native to the language is better here

pkg/optimizelyjson/optimizely_json_test.go
```diff
@@ -0,0 +1,198 @@
...
+func (suite *OptimizelyJsonTestSuite) TestGetValueInvalidJsonMultipleKeyEmptySchema() {
+	emptyStruct := struct{}{}
+	err := suite.optimizelyJson.GetValue("field3.some_key", &emptyStruct)
+	suite.Error(err)
+	suite.Equal(`json key "some_key" not found`, err.Error())
+	suite.Equal(struct{}{}, emptyStruct)
+}
+
+func (suite *OptimizelyJsonTestSuite) TestGetValueValidJsonKeyEmptySchema() {
+	emptyStruct := struct{}{}
```
msohailhussain, line 60: Am unable to understand, `emptyStruct` is empty but when I call `GetValue` with field4, shouldn't it set `field4` data.

pkg/optimizelyjson/optimizely_json.go
```diff
@@ -0,0 +1,98 @@
...
+	return string(jsonBytes), nil
+}
+
+// ToDict returns the native representation of json (map of interface)
+func (optlyJson OptimizelyJSON) ToDict() map[string]interface{} {
+	return optlyJson.jsonRepr
+}
+
+// GetValue populates the schema passed by the user - it takes primitive types and complex struct type
+func (optlyJson OptimizelyJSON) GetValue(jsonKey string, schema interface{}) error {
```
msohailhussain, line 53: Shouldn't it be `SetValue` since it's not returning anything. I will also recommend to use `jsonPath` instead of `jsonKey`

msohailhussain: `schema` is about metadata but it's all `data` mapped with the `jsonPath`. I think `schema` term doesn't fit here.

pkg/optimizelyjson/optimizely_json_test.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (suite *OptimizelyJsonTestSuite) SetupTest() {
+
+	suite.dynamicList = []interface{}{"1", "2", 3.01, 4.23, true}
+	suite.jsonRepr = map[string]interface{}{
+		"field1": 1,
+		"field2": 2.5,
+		"field3": "three",
+		"field4": map[string]interface{}{"inner_field1": 3, "inner_field2": suite.dynamicList},
```
msohailhussain, line 22: Can you please add a test case, how I can access `dynamicList`'s element. 
Ideally it should be `field4.inner_field2[2]` but I don't see `index` implementation for `GetValue` method.

pkg/optimizelyjson/optimizely_json.go
```diff
@@ -0,0 +1,98 @@
...
+
+// OptimizelyJSON holds the underlying structure of the object
+type OptimizelyJSON struct {
+	jsonRepr map[string]interface{}
+}
+
+// NewOptimizelyJSON constructs the object
+func NewOptimizelyJSON(jsonRepr map[string]interface{}) *OptimizelyJSON {
+	return &OptimizelyJSON{jsonRepr: jsonRepr}
+}
```
mikecdavis, line 35: We haven't tied in the project config yet, but I think this should probably be instantiated with `payload string` since that is what we are storing in the datafile. Then, depending on the method calls below we'd, memoize the transformations into `map[string]interface{}` if that was ever requested. As it's written now I feel we'd be marshaling data back and forth unnecessarily.

pkg/optimizelyjson/optimizely_json.go
mikecdavis: Agree `ToMap` makes more sense in this context.

pkg/optimizelyjson/optimizely_json.go
pawels-optimizely: > We haven't tied in the project config yet, but I think this should probably be instantiated with `payload string` since that is what we are storing in the datafile. Then, depending on the method calls below we'd, memoize the transformations into `map[string]interface{}` if that was ever requested. As it's written now I feel we'd be marshaling data back and forth unnecessarily.

I thought about that too, but then I thought that 2 out of 3 methods need map[string]interface{}, and only one method needs it as a string. But I agree that public SDK's would need to take the string and Unmarshal into that interface.

pkg/optimizelyjson/optimizely_json_test.go
pawels-optimizely: maybe that test is not needed at all, I wanted to show that there is no error even from passing empty schema.

pkg/optimizelyjson/optimizely_json_test.go
pawels-optimizely: Sure, I will provide that test, although officially we won't support array object, but that does not prevent the user to include array in json .

pkg/optimizelyjson/optimizely_json_test.go
msohailhussain: got it. already found a test where we are asserting partial and full schema.

pkg/optimizelyjson/optimizely_json.go
pawels-optimizely: I have implemented string payload, with map[string]interface{} used for memoization. I did look at the reflect package and wanted to avoid this memoization. Things get really complicated when the user does not provide a json tag. Unmarshal() can still match by field name, case insensitive. That would require creating a map of field names or something like that (steal some logic from Unmarshal()...), which I don't think will be that pretty if you want to preserve that functionality.  Also, I don't think we can enforce the user to provide json tags all the time.

PR #251, opened 2020-04-07
feat: added GetFeatureVariableJSON and GetAllFeatureVariables

pawels-optimizely:
## Summary
- renamed GetAllFeatureVariables to GetAllFeatureVariablesWithDecision - still needed for Agent
- fixed error handling in GetAllFeatureVariablesWithDecision
- added 2 public API's GetFeatureVariableJSON and GetAllFeatureVariables

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -18,7 +18,6 @@
 package client
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -227,12 +226,11 @@ func (o *OptimizelyClient) GetFeatureVariableJSON(featureKey, variableKey string
 		return value, err
 	}
 
-	var convertedValue map[string]interface{}
-	err = json.Unmarshal([]byte(val), &convertedValue)
+	value, err = optimizelyjson.NewOptimizelyJSONfromString(val)
 	if err != nil || valueType != entities.JSON {
-		return value, fmt.Errorf("variable value for key %s is invalid or wrong type", variableKey)
+		return nil, fmt.Errorf("variable value for key %s is invalid or wrong type", variableKey)
 	}
-	value = optimizelyjson.NewOptimizelyJSONfromMap(convertedValue)
+
 	return value, err
 }
 
@@ -299,9 +297,8 @@ func (o *OptimizelyClient) GetAllFeatureVariablesWithDecision(featureKey string,
 			out, err = strconv.Atoi(val)
 			errs = multierror.Append(errs, err)
 		case entities.JSON:
-			var intf map[string]interface{}
-			err = json.Unmarshal([]byte(val), &intf)
-			out = intf
+			var optlyJSON, err = optimizelyjson.NewOptimizelyJSONfromString(val)
+			out = optlyJSON.ToMap()
 			errs = multierror.Append(errs, err)
 		case entities.String:
 		default:
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -18,6 +18,7 @@
 package client
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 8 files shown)


pkg/client/client.go
```diff
@@ -215,6 +219,23 @@ func (o *OptimizelyClient) GetFeatureVariableString(featureKey, variableKey stri
...
+// GetFeatureVariableJSON returns the feature variable value of type json associated with the given feature and variable keys.
+func (o *OptimizelyClient) GetFeatureVariableJSON(featureKey, variableKey string, userContext entities.UserContext) (value *optimizelyjson.OptimizelyJSON, err error) {
+
+	val, valueType, err := o.GetFeatureVariable(featureKey, variableKey, userContext)
+	if err != nil {
+		return value, err
+	}
+
+	var convertedValue map[string]interface{}
+	err = json.Unmarshal([]byte(val), &convertedValue)
```
mikecdavis, line 231: We're swallowing this `err` here. That said we don't have to `Unmarshal` here. Why not just pass the to the optimizelyjson `string` constructor and let that object do it's job?

pkg/client/client.go
```diff
@@ -254,6 +275,8 @@ func (o *OptimizelyClient) GetAllFeatureVariables(featureKey string, userContext
 		return enabled, variableMap, nil
 	}
 
+	errs := new(multierror.Error)
```
mikecdavis, line 278: This is an improvement since before we only propagated the _last_ error. Thanks.

pkg/client/client.go
```diff
@@ -268,10 +291,18 @@ func (o *OptimizelyClient) GetAllFeatureVariables(featureKey string, userContext
...
 		case entities.Double:
 			out, err = strconv.ParseFloat(val, 64)
+			errs = multierror.Append(errs, err)
 		case entities.Integer:
 			out, err = strconv.Atoi(val)
+			errs = multierror.Append(errs, err)
+		case entities.JSON:
+			var intf map[string]interface{}
+			err = json.Unmarshal([]byte(val), &intf)
+			out = intf
```
mikecdavis, line 304: To avoid this `Unmarshal` being implemented in multiple places we can use the OptimizelyJSON object here.
```go
var json = optimizelyjson.NewOptimizelyJSONfromString(val)
out, err = json.ToMap()
```

pkg/client/client.go
pawels-optimizely: we don't swallow this error, we still check for this error and return the same message as for other variables. However, this thing will be rewritten to pass a string as the parameter.

pkg/client/client.go
```diff
@@ -227,12 +226,11 @@ func (o *OptimizelyClient) GetFeatureVariableJSON(featureKey, variableKey string
...
 
-	var convertedValue map[string]interface{}
-	err = json.Unmarshal([]byte(val), &convertedValue)
+	value, err = optimizelyjson.NewOptimizelyJSONfromString(val)
 	if err != nil || valueType != entities.JSON {
-		return value, fmt.Errorf("variable value for key %s is invalid or wrong type", variableKey)
+		return nil, fmt.Errorf("variable value for key %s is invalid or wrong type", variableKey)
 	}
-	value = optimizelyjson.NewOptimizelyJSONfromMap(convertedValue)
+
```
mikecdavis: Perhaps be:
```go
if valueType != entities.JSON {
	return nil, fmt.Errorf("variable value for key %s is invalid or wrong type", variableKey)
}

return optimizelyjson.NewOptimizelyJSONfromString(val)
```

pkg/optimizelyjson/optimizely_json_test.go
```diff
@@ -26,12 +26,13 @@ func (suite *OptimizelyJsonTestSuite) SetupTest() {
 		"field5": true,
 		"field6": nil,
 	}
-	suite.optimizelyJson = NewOptimizelyJSONfromString(suite.payload)
+	suite.optimizelyJson, _ = NewOptimizelyJSONfromString(suite.payload)
```
mikecdavis, line 29: `suite.NoError(err)`?

pkg/client/client.go
pawels-optimizely: since we have the same message for all variables ("invalid or wrong type") then the variable validation has to happen first.

PR #253, opened 2020-04-21
feat: added notification support for json variable

pawels-optimizely:
## Summary
- properly handling json variable in the notification listener

### DIFF
--- a/pkg/decision/composite_service.go
+++ b/pkg/decision/composite_service.go
@@ -18,6 +18,7 @@
 package decision
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


pkg/decision/composite_service.go
```diff
@@ -113,6 +114,9 @@ func (s CompositeService) GetFeatureDecision(featureDecisionContext FeatureDecis
 				convertedValue, e = strconv.ParseFloat(variableValue, 64)
 			case entities.Boolean:
 				convertedValue, e = strconv.ParseBool(variableValue)
+			case entities.JSON:
+				convertedValue = map[string]string{}
+				e = json.Unmarshal([]byte(variableValue), &convertedValue)
```
mikeng13, line 119: I'm wondering now, if we should lazy unmarshal and cache the map in the variable itself the first time it is ever accessed and from there on we can just refer to that value. Seems like this could potentially be a costly operation to repeat

mjc1283: Good point

mjc1283, line 119: Should we pass the OptimizelyJSON object to the notification, rather than the JSON string? And apply memoization to creating the OptimizelyJSON as @mikeng13 suggested?

pawels-optimizely: memoization is already done on OptimizelyJSON, the thing is that GetFeatureDecision does not know anything about OptimizelyJSON. This whole notification logic will need to be restructured.

PR #254, opened 2020-04-22
fix(httpDispatcher): bug. No logger passed in causing a crash.

thomaszurkan-optimizely:


pkg/event/dispatcher.go
```diff
@@ -65,6 +65,17 @@ func (ed *HTTPEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
 	return success, err
 }
 
+func NewHttpEventDispatcher(sdkKey string, requester *utils.HTTPRequester, logger logging.OptimizelyLogProducer) Dispatcher {
```
mikeng13, line 68: Need to be consistent with naming and make `HTTP` all caps here since you have it for the requester factory method below

PR #257, opened 2020-04-24
feat: changes in feature decision notification

pawels-optimizely:
## Summary
- move feature notification with variables to the client in order to reuse already converted variables
- although feature notification with variables was moved to the client, the registry still happens in the old way - on decision service. This is done this way not to cause any breaking changes.
- added all-feature-variable notification type
- cleaned up client tests


pkg/client/client.go
```diff
@@ -120,6 +121,14 @@ func (o *OptimizelyClient) IsFeatureEnabled(featureKey string, userContext entit
 		o.logger.Info(fmt.Sprintf(`Feature "%s" is not enabled for user "%s".`, featureKey, userContext.ID))
 	}
 
+	if o.notificationCenter != nil {
+		decisionNotification := decision.FeatureNotification(featureKey, &featureDecision, &userContext)
+
+		if e := o.notificationCenter.Send(notification.Decision, *decisionNotification); e != nil {
+			o.logger.Warning("Problem with sending notification")
```
mjc1283, line 128: Should this be an error-level log, instead of warning?

pkg/client/client.go
```diff
@@ -180,77 +209,191 @@ func (o *OptimizelyClient) GetFeatureVariableBoolean(featureKey, variableKey str
...
 
 // GetFeatureVariableJSON returns the feature variable value of type json associated with the given feature and variable keys.
 func (o *OptimizelyClient) GetFeatureVariableJSON(featureKey, variableKey string, userContext entities.UserContext) (value *optimizelyjson.OptimizelyJSON, err error) {
 
-	val, valueType, err := o.GetFeatureVariable(featureKey, variableKey, userContext)
+	val, valueType, featureDecision, err := o.getFeatureVariable(featureKey, variableKey, userContext)
+	defer func() {
+		if o.notificationCenter != nil {
+			var variableValue interface{}
+			if value != nil {
```
mjc1283, line 315: When would `value` be `nil`? I thought we only allowed JSON variables to be objects at the top level?

pkg/client/client.go
```diff
@@ -180,77 +209,191 @@ func (o *OptimizelyClient) GetFeatureVariableBoolean(featureKey, variableKey str
...
 	}
+
 	return value, err
 }
 
 // GetFeatureVariableJSON returns the feature variable value of type json associated with the given feature and variable keys.
 func (o *OptimizelyClient) GetFeatureVariableJSON(featureKey, variableKey string, userContext entities.UserContext) (value *optimizelyjson.OptimizelyJSON, err error) {
 
-	val, valueType, err := o.GetFeatureVariable(featureKey, variableKey, userContext)
+	val, valueType, featureDecision, err := o.getFeatureVariable(featureKey, variableKey, userContext)
```
mjc1283, line 311: There are a lot of similar identifiers in this function: `val`, `value`, `variableValue`. Any more descriptive names would be helpful.

pkg/client/client.go
```diff
@@ -166,11 +175,31 @@ func (o *OptimizelyClient) GetEnabledFeatures(userContext entities.UserContext)
 // GetFeatureVariableBoolean returns the feature variable value of type bool associated with the given feature and variable keys.
 func (o *OptimizelyClient) GetFeatureVariableBoolean(featureKey, variableKey string, userContext entities.UserContext) (value bool, err error) {
 
-	val, valueType, err := o.GetFeatureVariable(featureKey, variableKey, userContext)
+	var convertedValue bool
+	val, valueType, featureDecision, err := o.getFeatureVariable(featureKey, variableKey, userContext)
+	defer func() {
```
mjc1283, line 180: Nit: I would find this easier to read without a `defer`. It's referencing values that are assigned below, and above, so I have to read both directions, whereas if it was just moved down and didn't use `defer`, I only have to read one direction.

pkg/client/client.go
```diff
@@ -180,77 +209,191 @@ func (o *OptimizelyClient) GetFeatureVariableBoolean(featureKey, variableKey str
...
-		return nil, fmt.Errorf("variable value for key %s is invalid or wrong type", variableKey)
+		value, err = nil, fmt.Errorf("variable value for key %s is invalid or wrong type", variableKey)
 	}
 
 	return value, err
 }
 
 // GetFeatureVariable returns feature variable as a string along with it's associated type.
-func (o *OptimizelyClient) GetFeatureVariable(featureKey, variableKey string, userContext entities.UserContext) (value string, valueType entities.VariableType, err error) {
+func (o *OptimizelyClient) getFeatureVariable(featureKey, variableKey string, userContext entities.UserContext) (string, entities.VariableType, *decision.FeatureDecision, error) {
```
mjc1283, line 346: Nit: Any more descriptive names we can use, rather than having `getFeatureVariable` and `GetFeatureVariable`? If not changing, the doc comment on this one needs fixing (capitalized).

pkg/client/client.go
```diff
@@ -180,77 +209,191 @@ func (o *OptimizelyClient) GetFeatureVariableBoolean(featureKey, variableKey str
...
-			return v.Value, variable.Type, err
+			return v.Value, variable.Type, &featureDecision, nil
 		}
 	}
 
-	return variable.DefaultValue, variable.Type, err
+	return variable.DefaultValue, variable.Type, &featureDecision, nil
+}
+
+// GetFeatureVariable returns feature variable as a string along with it's associated type.
```
mjc1283, line 364: This returns the converted value, not variable as a string.

pkg/client/client.go
pawels-optimizely: this is what it was before. Actually I was thinking to remove at all, as we have registry injection for this, and I cannot think of the scenario where it will become an error

pkg/client/client.go
pawels-optimizely: OptimizelyJSON is the object and it can be nil, since I return the pointer to the object.  GetFeatureVariableJSON returns *optimizelyjson.OptimizelyJSON

pkg/client/client.go
pawels-optimizely: I will see if I can come up with the better names

pkg/client/client.go
pawels-optimizely: defer makes it easier to make fully backward compatible with the previous version sent from composite_service. We have 2 returns before we get to the final one. if we remove defer then we need to send notification for each return.

pkg/client/client.go
pawels-optimizely: yeah, that was my bad, the comment should not belong here, thanks for catching. GetFeatureVariable has to stay here, since this is a public API. getFeatureVariable was done only for notification support - to separate this notification and not to have double calls. I did not want to add anything to this function name like getFeatureVariableWithNotification or getFeatureVariableHelper, but any suggestions are welcome.

pkg/client/client.go
pawels-optimizely: good catch, actually that was old definition and I have not changed it. I will fix it.

pkg/client/client.go
mjc1283: If it is `ni`l, is that an unexpected error condition, or does it happen expectedly in some case(s)?

pkg/client/client.go
mjc1283: I don't have any suggestions, but will let you know if I think of one.

pkg/client/client.go
pawels-optimizely: there is only one scenario where this becomes nil - when it cannot unmarshal properly. Since datafiles are providing all the variables as quoted strings, we need to unmarshal this first. If we cannot for some reason, then it's returning nil json object. So this is an expected error condition.

pkg/client/client.go
mikeng13: I think warning is fine as this isn't severe enough (catastrophic) to be an error

PR #259, opened 2020-05-04
add readme-sync integration

juancarlostong:
@fscelliott i significantly improved our original PoC for readme-sync integration, now instead of triggering builds in external repos, i bring the readme-sync2 repo to the go-sdk build and run everything locally. ie. we are no longer dependant on the script/run_readme_sync.sh script (in readme-sync2) anymore, we can just get rid of that.

this change makes it a lot more robust since it has many fewer moving parts and it's actually faster because we end up cloning a lot less stuff.

one important thing to note is that this stage will only work if the docs/ directory (and the markdown files inside) are already existing in the repo so some other PR that adds these will need to be merged ahead of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,7 +116,7 @@ jobs:
         #
         # first we list all possible sdks and inside the for loop, remove the one we are updating
         - export ALL_SDK_REPOS="android-sdk csharp-sdk go-sdk java-sdk javascript-sdk objective-c-sdk python-sdk react-sdk ruby-sdk swift-sdk"
-        - mkdir $HOME/sdks && pushd $HOME/sdks && for i in ${ALL_SDKS//${TRAVIS_REPO_SLUG#optimizely/}}; do git clone https://github.com/optimizely/$i; [ -d "$i/docs/readme-sync" ] && ln -s $i/docs/readme-sync/sdk-reference-guides/$i $HOME/readme-sync2/docs/readme-sync/sdk-reference-guides/$i; done && popd
+        - mkdir $HOME/sdks && pushd $HOME/sdks && for i in ${ALL_SDK_REPOS//${TRAVIS_REPO_SLUG#optimizely/}}; do git clone https://github.com/optimizely/$i; [ -d "$i/docs/readme-sync" ] && ln -s $i/docs/readme-sync/sdk-reference-guides/$i $HOME/readme-sync2/docs/readme-sync/sdk-reference-guides/$i; done && popd
         # check our work
         - ls -al $HOME/sdks
         - ls -al $HOME/readme-sync2/docs/readme-sync/sdk-reference-guides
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,11 +1,12 @@
 language: minimal
 env: GO111MODULE=on
 git:
-  depth: 1
+  depth: false
```
juancarlostong, line 4: @mikeng13 i had to set git.depth to false in order to run `git diff`

.travis.yml
```diff
@@ -19,8 +20,6 @@ jobs:
       install: gem install awesome_bot
       script:
         - find . -type f -name '*.md' -exec awesome_bot {} \;
-      notifications:
-        email: false
```
juancarlostong, line 23: this actually doesn't do anything, maybe it was feature that was deprecated a while back, so removing for clarity

.travis.yml
```diff
@@ -1,11 +1,12 @@
 language: minimal
 env: GO111MODULE=on
 git:
-  depth: 1
+  depth: false
 install:
   - eval "$(gimme)"
 stages:
-  - name: 'Lint markdown files'
```
juancarlostong, line 8: no real reason to specify 'name:' here so removing for consistency with other stages

.travis.yml
```diff
@@ -96,3 +95,23 @@ jobs:
...
+        - mkdir $HOME/readme-sync2 && pushd $HOME/readme-sync2 && git init && git pull https://$CI_USER_TOKEN@github.com/optimizely/readme-sync2.git && popd
+        - source ~/.nvm/nvm.sh && cd $HOME/readme-sync2 && nvm install && npm install
+        # this preps the input directory for readme-sync script
+        - mkdir -p $HOME/readme-sync2/docs/sdk-reference-guides
+        # ${TRAVIS_REPO_SLUG#optimizely/} translates to go-sdk
+        - ln -s $TRAVIS_BUILD_DIR/docs $HOME/readme-sync2/docs/sdk-reference-guides/${TRAVIS_REPO_SLUG#optimizely/}
+      script:
+        # we need to be in $TRAVIS_BUILD_DIR in order to run the following git diff properly
+        - cd $TRAVIS_BUILD_DIR
+        - git diff --quiet $TRAVIS_COMMIT_RANGE -- docs/ || ( cd $HOME/readme-sync2 && npx ts-node sync/index.ts --apiKey $README_SYNC_API_KEY --version 4.0 --docs docs/ )
```
juancarlostong, line 117: this is the stage that can be dropped into any *-sdk repo (as long as they have the docs/ folder already in place)
we'll also need to include $README_SYNC_API_KEY in the travis settings of every repo that wants to use this stage

juancarlostong:
we cant merge this because of some new issues we discovered, let's talk more and figure out direction first

juancarlostong:
this branch now contains a readme-sync stage that can be dropped into any other *-sdk repo to enable readme-sync. we should ensure that all the other sdks already have the docs/readme-sync/sdk-reference-guides/*-sdk folder before we try to sync because if they are omitted, syncing without it means we are going to delete the existing entry.

PR #266, opened 2020-06-01
feat: Added GetDetailedFeatureDecisionUnsafe

msohailhussain:
### Summary
Temporary changes in go-sdk for adding a new method GetDetailedFeatureDecisionUnsafe
This is added to support some agent features. Which will be removed later on.

### Testplan
- All FSC should be passed.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -430,9 +430,9 @@ func (o *OptimizelyClient) GetAllFeatureVariablesWithDecision(featureKey string,
 			}
 		}
 
-		out, typedError := o.getTypedValue(val, v)
+		typedValue, typedError := o.getTypedValue(val, v.Type)
 		errs = multierror.Append(errs, typedError)
-		variableMap[v.Key] = out
+		variableMap[v.Key] = typedValue
 	}
 
 	if o.notificationCenter != nil {
@@ -462,14 +462,16 @@ func (o *OptimizelyClient) GetDetailedFeatureDecisionUnsafe(featureKey string, u
 
 	if featureDecision.Variation != nil {
 		decisionInfo.Enabled = featureDecision.Variation.FeatureEnabled
-		decisionInfo.VariationKey = featureDecision.Variation.Key
-		decisionInfo.ExperimentKey = featureDecision.Experiment.Key
 
-		// Triggers impression events when applicable
-		if !disableTracking && featureDecision.Source == decision.FeatureTest {
-			// send impression event for feature tests
-			impressionEvent := event.CreateImpressionUserEvent(decisionContext.ProjectConfig, featureDecision.Experiment, *featureDecision.Variation, userContext)
-			o.EventProcessor.ProcessEvent(impressionEvent)
+		if featureDecision.Source == decision.FeatureTest {
+			decisionInfo.VariationKey = featureDecision.Variation.Key
+			decisionInfo.ExperimentKey = featureDecision.Experiment.Key
+			// Triggers impression events when applicable
+			if !disableTracking {
+				// send impression event for feature tests
+				impressionEvent := event.CreateImpressionUserEvent(decisionContext.ProjectConfig, featureDecision.Experiment, *featureDecision.Variation, userContext)
+				o.EventProcessor.ProcessEvent(impressionEvent)
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/client/client.go
```diff
@@ -467,6 +447,66 @@ func (o *OptimizelyClient) GetAllFeatureVariablesWithDecision(featureKey string,
...
+	for _, v := range feature.VariableMap {
+		val := v.DefaultValue
+
+		if decisionInfo.Enabled {
+			if variable, ok := featureDecision.Variation.Variables[v.ID]; ok {
+				val = variable.Value
+			}
+		}
+
+		out, typedError := o.getTypedValue(val, v)
```
mjc1283, line 493: Nit: `out` isn't an ideal variable name IMO. What does it mean?

pkg/client/client.go
```diff
@@ -467,6 +447,66 @@ func (o *OptimizelyClient) GetAllFeatureVariablesWithDecision(featureKey string,
...
+
+	errs := new(multierror.Error)
+
+	for _, v := range feature.VariableMap {
+		val := v.DefaultValue
+
+		if decisionInfo.Enabled {
+			if variable, ok := featureDecision.Variation.Variables[v.ID]; ok {
+				val = variable.Value
+			}
```
mjc1283, line 490: If `ok` is false, let's log an error.

pkg/client/client.go
```diff
@@ -467,6 +447,66 @@ func (o *OptimizelyClient) GetAllFeatureVariablesWithDecision(featureKey string,
...
+		decisionInfo.VariableMap[v.Key] = out
+	}
+
+	if o.notificationCenter != nil {
+		decisionNotification := decision.FeatureNotificationWithVariables(featureKey, &featureDecision, &userContext,
+			map[string]interface{}{"variableValues": decisionInfo.VariableMap})
+		decisionNotification.Type = notification.AllFeatureVariables
+
+		if err = o.notificationCenter.Send(notification.Decision, *decisionNotification); err != nil {
+			o.logger.Warning("Problem with sending notification")
```
mjc1283, line 504: Let's log `err` here so the user can see what the problem was.

pkg/client/client.go
```diff
@@ -685,6 +725,26 @@ func (o *OptimizelyClient) RemoveOnTrack(id int) error {
 	return nil
 }
 
+func (o *OptimizelyClient) getTypedValue(value string, variable entities.Variable) (out interface{}, err error) {
+	out = value
+	switch varType := variable.Type; varType {
```
mjc1283, line 730: Why do we need to use a new variable `varType`? Does this not work?
```suggestion
	switch variable.Type {
```

pkg/client/client.go
```diff
@@ -467,6 +447,66 @@ func (o *OptimizelyClient) GetAllFeatureVariablesWithDecision(featureKey string,
...
+	decisionContext, featureDecision, err := o.getFeatureDecision(featureKey, "", userContext)
+	if err != nil {
+		o.logger.Error("Optimizely SDK tracking error", err)
+		return decisionInfo, err
+	}
+
+	if featureDecision.Variation != nil {
+		decisionInfo.Enabled = featureDecision.Variation.FeatureEnabled
+		decisionInfo.VariationKey = featureDecision.Variation.Key
+		decisionInfo.ExperimentKey = featureDecision.Experiment.Key
```
mjc1283, line 466: I think we should only assign these when `featureDecision.Source == decision.FeatureTest`. Keys from a rollout experiment should not be exposed.

pkg/client/client.go
```diff
@@ -685,6 +725,26 @@ func (o *OptimizelyClient) RemoveOnTrack(id int) error {
 	return nil
 }
 
+func (o *OptimizelyClient) getTypedValue(value string, variable entities.Variable) (out interface{}, err error) {
```
pawels-optimizely, line 728: do we need to pass the whole variable ? Seems like only type will be enough.

pkg/client/client_test.go
```diff
@@ -1679,6 +1679,294 @@ func TestGetAllFeatureVariablesWithDecisionWithoutFeature(t *testing.T) {
...
+	testFeatureKey := "test_feature_key"
+	testUserContext := entities.UserContext{ID: "test_user_1"}
+
+	type variable struct {
+		key        string
+		defaultVal string
+		varVal     string
+		varType    entities.VariableType
+		expected   interface{}
+	}
```
mjc1283, line 1789: The same struct type is declared in the previous test. I suggest declaring it only once inside this file.

pkg/client/client_test.go
```diff
@@ -1679,6 +1679,294 @@ func TestGetAllFeatureVariablesWithDecisionWithoutFeature(t *testing.T) {
...
+
+		variableMap[id] = entities.Variable{
+			DefaultValue: v.defaultVal,
+			ID:           id,
+			Key:          v.key,
+			Type:         v.varType,
+		}
+
+		mockConfig.On("GetVariableByKey", testFeatureKey, v.key).Return(v.varVal, nil)
+	}
```
mjc1283, line 1722: This logic is repeated in several tests, please factor this out into a helper function.

PR #267, opened 2020-06-01
feat: added authenticated datafile support

pawels-optimizely:
## Summary
- added support for authenticated datafiles
- refactored static config manager  to look more like polling config manager
- fixing a bug for polling interval set to 0 or below.
- here are different ways of initializing authenticated datafiles:
```
/************* default client  polling mgr ********************/
optimizelyClient, _ := optimizely.Client(sdkKey, client.WithDatafileAccessToken(token))
optimizelyClient.IsFeatureEnabled("EVENT_BATCHING", user)

/*********** custom polling mgr *****************/
pcm := config.NewPollingProjectConfigManager(sdkKey, config.WithDatafileAccessToken(token))
optimizelyClient, _ = optimizely.Client(sdkKey, client.WithConfigManager(pcm))
optimizelyClient.IsFeatureEnabled("EVENT_BATCHING", user)

/*************** custome polling mgr , setting polling interval *********/
optimizelyClient, _ = optimizely.Client(sdkKey, client.WithPollingConfigManagerDatafileAccessToken(time.Second, nil, token),
client.WithBatchEventProcessor(event.DefaultBatchSize, event.DefaultEventQueueSize, event.DefaultEventFlushInterval))
optimizelyClient.IsFeatureEnabled("EVENT_BATCHING", user)

/*************** static client *********/
optimizelyFactory := &client.OptimizelyFactory{
	SDKKey:              sdkKey,
	DatafileAccessToken: token,
}

optimizelyClient, _ = optimizelyFactory.StaticClient()
optimizelyClient.IsFeatureEnabled("EVENT_BATCHING", user)

/******** custom static client init ***************/
cm := config.NewStaticProjectConfigManager(sdkKey, config.WithDatafileAccessToken(token))
optimizelyClient, _ = optimizely.Client(sdkKey, client.WithConfigManager(cm))
optimizelyClient.IsFeatureEnabled("EVENT_BATCHING", user)

```

### DIFF
--- a/pkg/client/factory_test.go
+++ b/pkg/client/factory_test.go
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2019, Optimizely, Inc. and contributors                        *
+ * Copyright 2019-2020, Optimizely, Inc. and contributors                   *
  *                                                                          *
  * Licensed under the Apache License, Version 2.0 (the "License");          *
  * you may not use this file except in compliance with the License.         *
--- a/pkg/client/factory_test.go
+++ b/pkg/client/factory_test.go
@@ -192,10 +192,12 @@ func TestClientMetrics(t *testing.T) {
 
 func TestClientWithDatafileAccessToken(t *testing.T) {
 	factory := OptimizelyFactory{SDKKey: "1212"}
-
-	optimizelyClient, err := factory.Client(WithDatafileAccessToken("some_token"))
+	accessToken := "some_token"
+	optimizelyClient, err := factory.Client(WithDatafileAccessToken(accessToken))
 	assert.NoError(t, err)
 	assert.NotNil(t, optimizelyClient.ConfigManager)
 	assert.NotNil(t, optimizelyClient.DecisionService)
 	assert.NotNil(t, optimizelyClient.EventProcessor)
+
+	assert.Equal(t, accessToken, factory.DatafileAccessToken)
 }
--- a/pkg/client/factory_test.go
+++ b/pkg/client/factory_test.go
@@ -25,10 +25,8 @@ import (
 	"time"
 
 	"github.com/optimizely/go-sdk/pkg/config"
-	"github.com/optimizely/go-sdk/pkg/config/datafileprojectconfig"
 	"github.com/optimizely/go-sdk/pkg/decision"
 	"github.com/optimizely/go-sdk/pkg/event"
-	"github.com/optimizely/go-sdk/pkg/logging"
 	"github.com/optimizely/go-sdk/pkg/metrics"
 	"github.com/optimizely/go-sdk/pkg/utils"
 
@@ -85,10 +83,19 @@ func TestClientWithPollingConfigManager(t *testing.T) {
 	assert.NotNil(t, optimizelyClient.EventProcessor)
 }
 
+func TestClientWithPollingConfigManagerDatafileAccessToken(t *testing.T) {
+	factory := OptimizelyFactory{}
+
+	optimizelyClient, err := factory.Client(WithPollingConfigManagerDatafileAccessToken(time.Hour, nil, "some_token"))
+	assert.NoError(t, err)
+	assert.NotNil(t, optimizelyClient.ConfigManager)
+	assert.NotNil(t, optimizelyClient.DecisionService)
+	assert.NotNil(t, optimizelyClient.EventProcessor)
+}
 func TestClientWithProjectConfigManagerInOptions(t *testing.T) {
 	factory := OptimizelyFactory{}
-	projectConfig := datafileprojectconfig.DatafileProjectConfig{}
-	configManager := config.NewStaticProjectConfigManager(projectConfig, logging.GetLogger("", ""))
+	mockDatafile := []byte(`{"version":"4"}`)
+	configManager := config.NewStaticProjectConfigManager("", config.WithInitialDatafile(mockDatafile))
 
 	optimizelyClient, err := factory.Client(WithConfigManager(configManager))
 	assert.NoError(t, err)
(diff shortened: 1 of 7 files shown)


pkg/config/polling_manager.go
```diff
@@ -182,48 +197,58 @@ func (cm *PollingProjectConfigManager) Start(ctx context.Context) {
...
-		sdkKey:              sdkKey,
-		logger:              logging.GetLogger(sdkKey, "PollingProjectConfigManager"),
-	}
-
-	for _, opt := range pollingMangerOptions {
-		opt(&pollingProjectConfigManager)
-	}
-
+	pollingProjectConfigManager := newConfigManager(sdkKey, logging.GetLogger(sdkKey, "PollingProjectConfigManager"), pollingMangerOptions...)
 	pollingProjectConfigManager.setInitialDatafile(pollingProjectConfigManager.initDatafile)
```
mjc1283, line 249: Unrelated to your PR: it is suspicious that, in `NewPollingProjectConfigManager`, this `setInitialDatafile` call is surrounded by `if len(pollingProjectConfigManager.initDatafile) > 0 {`, while here, it is not surrounded by an `if`. Do we need this check or not?

pkg/config/polling_manager_test.go
```diff
@@ -626,3 +626,54 @@ func TestWithRequester(t *testing.T) {
...
+
+	assert.Equal(t, configManagerRequester, configManager.requester)
+	assert.Equal(t, asyncConfigManagerRequester, asyncConfigManager.requester)
+
+	configManager.setDatafileAccessTokenRequest()
+	asyncConfigManager.setDatafileAccessTokenRequest()
+
+	// modified requester check
+	assert.NotEqual(t, configManagerRequester, configManager.requester)
+	assert.NotEqual(t, asyncConfigManagerRequester, asyncConfigManager.requester)
```
mjc1283, line 678: I think we also need an assertion that it sets the `Authorization: Bearer` header correctly in the request.

pkg/config/static_manager.go
```diff
@@ -19,56 +19,37 @@ package config
...
+func NewStaticProjectConfigManager(sdkKey string, configMangerOptions ...OptionFunc) *StaticProjectConfigManager {
 
 	logger := logging.GetLogger(sdkKey, "StaticProjectConfigManager")
-
-	url := fmt.Sprintf(DatafileURLTemplate, sdkKey)
-	datafile, _, code, e := requester.Get(url)
-	if e != nil {
-		logger.Error(fmt.Sprintf("request returned with http code=%d", code), e)
-		return nil, e
+	staticProjectConfigManager := newConfigManager(sdkKey, logger, configMangerOptions...)
```
mjc1283, line 39: Does `config.WithDatafileAccessToken` work with `NewStaticProjectConfigManager`? I think it should. Also that implies we need the logic to switch default URL template in here, probably refactored into a shared helper method.

pkg/config/polling_manager.go
```diff
@@ -182,48 +197,58 @@ func (cm *PollingProjectConfigManager) Start(ctx context.Context) {
...
-func NewPollingProjectConfigManager(sdkKey string, pollingMangerOptions ...OptionFunc) *PollingProjectConfigManager {
+func (cm *PollingProjectConfigManager) setDatafileAccessTokenRequest() {
+	if cm.datafileAccessToken != "" {
+		headers := []utils.Header{{Name: "Content-Type", Value: "application/json"}, {Name: "Accept", Value: "application/json"}}
+		headers = append(headers, utils.Header{Name: "Authorization", Value: "Bearer " + cm.datafileAccessToken})
+		cm.requester = utils.NewHTTPRequester(logging.GetLogger(cm.sdkKey, "HTTPRequester"), utils.Headers(headers...))
+	}
+}
 
+func newConfigManager(sdkKey string, logger logging.OptimizelyLogProducer, configOptions ...OptionFunc) *PollingProjectConfigManager {
```
mjc1283, line 208: Nit: consider putting `newConfigManager` in its own file. It is not specific to polling manager.

pkg/config/polling_manager.go
pawels-optimizely: we can, but we don't have to. I will make this change, but the end result will be the same. in NewPollingProjectConfigManager it is really if/else , here we don't have if/else and we try loading initial file first. if we have nil for initial file, then it will not load but we will see the error in the log. So the end result will be the same, but it will be cleaner to check it first. good catch

pkg/config/polling_manager.go
pawels-optimizely: it is kind of related to polling config manager, because newConfigManager is constructing polling config manager object that can be safely used by static config manager. It is just a helper method, getting all common things together

pkg/config/polling_manager_test.go
pawels-optimizely: I agree with your comment. However, we have no good way to test headers on the  polling manager requester, because we declared it as private. We could either make the requester public or add another API on requester (GetHeaders()), but I don't think that will be a good direction. We test header addition on requester separately

pkg/config/static_manager.go
pawels-optimizely: yes that works, also, if you initialize static manager like this:
```
cm := config.NewStaticProjectConfigManager(sdkKey, config.WithDatafileAccessToken(token), config.WithDatafileURLTemplate(config.DatafileURLTemplate))
```
for private datafile then it will not work

pkg/client/factory_test.go
```diff
@@ -182,3 +189,13 @@ func TestClientMetrics(t *testing.T) {
...
+
+func TestClientWithDatafileAccessToken(t *testing.T) {
+	factory := OptimizelyFactory{SDKKey: "1212"}
+
+	optimizelyClient, err := factory.Client(WithDatafileAccessToken("some_token"))
+	assert.NoError(t, err)
+	assert.NotNil(t, optimizelyClient.ConfigManager)
+	assert.NotNil(t, optimizelyClient.DecisionService)
+	assert.NotNil(t, optimizelyClient.EventProcessor)
+}
```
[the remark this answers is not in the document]
pawels-optimizely: easy change, done

PR #273, opened 2020-06-18
fix: provide synchronization for Notification Center Cache

pawels-optimizely:
## Summary
- provide synchronization for Notification Center Cache
- this fix is for a case when many clients are initialized with different sdk keys ( Agent scenario)


mikeng13:
Ah, should we add unit tests for it then?

PR #275, opened 2020-07-01
Fix unbounded memory allocation.

mikecdavis:
## Summary
* Add bounded capacity to `InMemoryQueue` directly
* Fix potential concurrency issues in `InMemoryQueue` when checking queue size
* Fix `QueueEventDispatcher` size gauge metric
* Fix leaking goroutines in `QueueEventDispatcher`
* Run `go mod tidy`

## Issues
* OASIS-6735
* OASIS-6738


pkg/event/queue.go
```diff
@@ -19,60 +19,85 @@ package event
...
 import (
 	"sync"
+
+	"github.com/optimizely/go-sdk/pkg/logging"
 )
 
 // Queue represents a queue
 type Queue interface {
-	Add(item interface{})
+	Add(item interface{}) // Should return a bool
```
aliabbasrizvi, line 28: TODO?

pkg/event/queue.go
```diff
@@ -19,60 +19,85 @@ package event
...
-	i.Mux.Lock()
-	i.Queue = append(i.Queue, item)
-	i.Mux.Unlock()
+func (q *InMemoryQueue) Add(item interface{}) {
+	q.Mux.Lock()
+	defer q.Mux.Unlock()
+
+	if len(q.Queue) >= q.MaxSize {
+		q.logger.Warning("MaxQueueSize has been met. Discarding event")
+		return
```
aliabbasrizvi, line 58: Shouldn't this be returning error or something? Or is that planned in a follow up?

pkg/event/queue_test.go
```diff
@@ -22,14 +22,25 @@ import (
 	"testing"
 )
 
+func TestInMemoryQueue_Add(t *testing.T) {
```
aliabbasrizvi, line 25: nit. This could probably be more descriptive i.e. `TestInMemoryQueue_Add_Size_Bounded_By_MaxSize`?

pkg/event/dispatcher.go
```diff
@@ -94,24 +95,23 @@ type QueueEventDispatcher struct {
 // DispatchEvent queues event with callback and calls flush in a go routine.
 func (ed *QueueEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
 	ed.eventQueue.Add(event)
-	go func() {
-		ed.flushEvents()
```
pawels-optimizely, line 98: duh, hehe

pkg/event/dispatcher.go
```diff
@@ -94,24 +95,23 @@ type QueueEventDispatcher struct {
...
+	if !ed.processing.TryAcquire(1) {
+		return
+	}
+	defer ed.processing.Release(1)
 
 	retryCount := 0
-	ed.queueSize.Set(float64(ed.eventQueue.Size()))
-	for ed.eventQueue.Size() > 0 {
+	queueSize := ed.eventQueue.Size()
+	for ; queueSize > 0; queueSize = ed.eventQueue.Size() {
```
pawels-optimizely, line 113: I think I understand, but can you provide a quick comment explaining why this has to be in the loop?

pkg/event/dispatcher.go
```diff
@@ -159,7 +158,7 @@ func (ed *QueueEventDispatcher) flushEvents() {
 			ed.retryFlushCounter.Add(1)
 		}
 	}
-	ed.queueSize.Set(float64(ed.eventQueue.Size()))
+	ed.queueSize.Set(float64(queueSize))
```
pawels-optimizely, line 161: shouldn't it get the right size all the time when we use ed.eventQueue.Size() ? (just asking)

pkg/event/dispatcher.go
```diff
@@ -136,7 +136,6 @@ func (ed *QueueEventDispatcher) flushEvents() {
 
 		if err == nil {
 			if success {
-				ed.logger.Debug(fmt.Sprintf("Dispatched log event %+v", event))
```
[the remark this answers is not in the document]
mikecdavis: This line is wasting resources. The event payload (which is large) is getting parsed into this string representation regardless of the log level. Our logging interface should really have a `Debugf(...)`, etc method for these types of formatted messages.

pkg/event/queue.go
mikecdavis: Yeah probably a TODO item, but unfortunately it would break the interface otherwise I would have done it here. Likely we'd have to define another interface and add backwards compatibility but wrapping any old implementations into the new one.

pkg/event/queue.go
mikecdavis: We should yes, and this is what prompted me to add that comment on the struct to return a boolean. So for now planned follow-up.

pkg/event/dispatcher.go
mikecdavis: `eventQueue.Size()` is a blocking call and I didn't feel it necessary to call it again here since it was just called from the for loop and we presumably exited with a size of zero. If the queue had changed since then it will be captured with the next call to `flushEvents()`.

pkg/event/dispatcher.go
mikecdavis: Since the Size() was only evaluated at the beginning and the end of the loop. We failed to capture any queue size that may have evolved while the loop was being processed. Under heavy load, the queue can grow and this loop will continue to iterate, but the size would not be captured. I discovered this while investigating runaway memory allocation. Our metrics were reporting a constant size in the single digits, but in reality it was growing into the hundreds of thousands.

pkg/event/dispatcher.go
```diff
@@ -172,13 +171,15 @@ func NewQueueEventDispatcher(sdkKey string, metricsRegistry metrics.Registry) *Q
...
-		eventQueue:        NewInMemoryQueue(defaultQueueSize),
+		eventQueue:        NewInMemoryQueueWithLogger(defaultQueueSize, logger),
 		Dispatcher:        NewHTTPEventDispatcher(sdkKey, nil, nil),
 		queueSize:         dispatcherMetricsRegistry.GetGauge(metrics.DispatcherQueueSize),
 		retryFlushCounter: dispatcherMetricsRegistry.GetCounter(metrics.DispatcherRetryFlush),
 		failFlushCounter:  dispatcherMetricsRegistry.GetCounter(metrics.DispatcherFailedFlush),
 		sucessFlush:       dispatcherMetricsRegistry.GetCounter(metrics.DispatcherSuccessFlush),
-		logger:            logging.GetLogger(sdkKey, "QueueEventDispatcher"),
+		logger:            logger,
+		processing:        semaphore.NewWeighted(int64(1)),
```
msohailhussain, line 183: can we make it `const`

pkg/event/dispatcher.go
```diff
@@ -94,24 +95,23 @@ type QueueEventDispatcher struct {
...
 // flush the events
 func (ed *QueueEventDispatcher) flushEvents() {
 
-	ed.eventFlushLock.Lock()
-
-	defer func() {
-		ed.eventFlushLock.Unlock()
-	}()
+	// Limit flushing to a single worker
+	if !ed.processing.TryAcquire(1) {
```
msohailhussain, line 106: to use that `const` value here.

pkg/event/dispatcher.go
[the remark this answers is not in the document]
mjc1283: Might want to keep this line, but don't parse the whole event. It can be useful.

PR #276, opened 2020-07-06
Update test timing to fix flaky assertion.

mikecdavis:
## Summary
* Move sleep until after counters are updated.
* Increase sleep within the test to be less flaky.

@thomaszurkan-optimizely I noticed that within this test it says:
> // check the queue. bad event type should be removed.  but, not sent.

But the event is _not_ removed and we're asserting on a non-empty queue. Is this a bug or should the comment be updated?


thomaszurkan-optimizely:
@mikecdavis not sure what is going on here. But, you removed the flush before the sleep so of course there should be no drain.  And, the mock dispatcher is set to fail.  So, it should still be there since the dispatcher failed.  Maybe there was a bad object added as well that should be removed?  It appears all this has been refactored and so I'm not sure anymore.

PR #280, opened 2020-07-14
refact(audience-logs): Added and refactored audience evaluation logs. (WIP)

yasirfolio3:
## Summary
- Added and updated missing audience evaluation logs.

## Issues
https://optimizely.atlassian.net/browse/OASIS-6816


pkg/decision/evaluator/audience.go
```diff
@@ -1,45 +0,0 @@
-/****************************************************************************
```
msohailhussain, line 1: Why deleted this file?

pkg/decision/evaluator/condition.go
```diff
@@ -38,14 +39,23 @@ type ItemEvaluator interface {
...
+}
 
 // Evaluate returns true if the given user's attributes match the condition
 func (c CustomAttributeConditionEvaluator) Evaluate(condition entities.Condition, condTreeParams *entities.TreeParameters) (bool, error) {
 	// We should only be evaluating custom attributes
 
 	if condition.Type != customAttributeType {
-		return false, fmt.Errorf(`unable to evaluator condition of type "%s"`, condition.Type)
+
+		c.logger.Warning(fmt.Sprintf(string(logging.UnknownConditionType), condition.StringRepresentation))
```
msohailhussain, line 57: Why string is needed? string(logging.UnknownConditionType)

pkg/decision/evaluator/condition.go
```diff
@@ -57,6 +67,7 @@ func (c CustomAttributeConditionEvaluator) Evaluate(condition entities.Condition
 	case exactMatchType:
 		matcher = matchers.ExactMatcher{
 			Condition: condition,
+			Logger:    c.logger,
```
msohailhussain, line 70: Why logger not passed in `existsMatchType`

pkg/decision/evaluator/condition.go
```diff
@@ -65,16 +76,20 @@ func (c CustomAttributeConditionEvaluator) Evaluate(condition entities.Condition
...
 		}
 	case substringMatchType:
 		matcher = matchers.SubstringMatcher{
 			Condition: condition,
+			Logger:    c.logger,
 		}
 	default:
+		c.logger.Warning(fmt.Sprintf(string(logging.UnknownMatchType), condition.StringRepresentation))
```
msohailhussain, line 92: Please use format specifier.

pkg/decision/evaluator/matchers/lt.go
```diff
@@ -18,27 +18,17 @@
...
-	if floatValue, ok := utils.ToFloat(m.Condition.Value); ok {
-		attributeValue, err := user.GetFloatAttribute(m.Condition.Name)
-		if err != nil {
-			return false, err
-		}
-		return floatValue > attributeValue, nil
-	}
-
-	return false, fmt.Errorf("audience condition %s evaluated to NULL because the condition value type is not supported", m.Condition.Name)
+	return matchGtOrLt(user, m.Condition, m.Logger, false)
```
msohailhussain, line 33: Why we didn't only add log here.

pkg/entities/user_context.go
```diff
@@ -88,6 +88,15 @@ func (u UserContext) GetIntAttribute(attrName string) (int64, error) {
 	return 0, fmt.Errorf(`no int attribute named "%s"`, attrName)
 }
 
+// GetAttribute returns the value for the specified attribute name in the attributes map. Returns error if not found.
+func (u UserContext) GetAttribute(attrName string) (interface{}, error) {
```
msohailhussain, line 92: Why added this method?

pkg/logging/log_messages.go
```diff
@@ -0,0 +1,60 @@
+/****************************************************************************
+ * Copyright 2019-2020, Optimizely, Inc. and contributors                   *
```
msohailhussain, line 2: Is it started in 2019?

pkg/logging/log_messages.go
```diff
@@ -0,0 +1,60 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package logging //
+package logging
+
+// LogMessage defines string type for log messages
+type LogMessage string
```
msohailhussain, line 21: I am not sure if this is the right location to add log messages. Let's discuss offline.

pkg/decision/evaluator/condition.go
msohailhussain, line 57: We don't really need two Strings. Just pass the same in fmt.Errorf

msohailhussain: I will confirm and let you know.

pkg/decision/evaluator/audience.go
yasirfolio3: this file wasn't being used anywhere throughout the sdk and was throwing syntax errors due to the changes made in this PR.

pkg/decision/evaluator/condition.go
yasirfolio3: No logs were required inside `exists`.

pkg/decision/evaluator/condition.go
yasirfolio3: Since `UnknownConditionType` is of custom type, we must convert it to string before formatting.

pkg/decision/evaluator/matchers/lt.go
yasirfolio3: since both `gt` and `lt` matchers were very similar, linter was throwing `lines are duplicate` error.

pkg/entities/user_context.go
yasirfolio3: for `InvalidAttributeValueType` log, we required the raw attribute value to log its original type.

pkg/decision/evaluator/matchers/exact.go
```diff
@@ -22,18 +22,28 @@ import (
...
 	Condition entities.Condition
+	Logger    logging.OptimizelyLogProducer
 }
 
 // Match returns true if the user's attribute match the condition's string value
 func (m ExactMatcher) Match(user entities.UserContext) (bool, error) {
+
+	if !user.CheckAttributeExists(m.Condition.Name) {
+		m.Logger.Debug(fmt.Sprintf(logging.NullUserAttribute.String(), m.Condition.StringRepresentation, m.Condition.Name))
+		return false, fmt.Errorf(`no attribute named "%s"`, m.Condition.Name)
```
msohailhussain, line 39: Why we don't return same logging message?

pkg/decision/evaluator/matchers/gt.go
```diff
@@ -22,18 +22,33 @@ import (
...
+	"github.com/optimizely/go-sdk/pkg/logging"
 )
 
 // GtMatcher matches against the "gt" match type
-func GtMatcher(condition entities.Condition, user entities.UserContext) (bool, error) {
+func GtMatcher(condition entities.Condition, user entities.UserContext, logger logging.OptimizelyLogProducer) (bool, error) {
+	return matchGtOrLt(condition, user, logger, true)
+}
+
+func matchGtOrLt(condition entities.Condition, user entities.UserContext, logger logging.OptimizelyLogProducer, gtMatch bool) (bool, error) {
```
mikecdavis, line 33: Nit, but If we make this a pure comparator and return `(int, bool)` e.g. -1, 0, 1, then we can drop the `gtMatch bool` which is a bit awkward. Then consumers would look like:
```go
func GtMatcher(condition entities.Condition, user entities.UserContext, logger logging.OptimizelyLogProducer) (bool, error) {
	res, err := compare(condition, user, logger)
	if err := nil {
		return false, err
	}
	return res > 0, nil
}

func LtMatcher(condition entities.Condition, user entities.UserContext, logger logging.OptimizelyLogProducer) (bool, error) {
	res, err := compare(condition, user, logger)
	if err := nil {
		return false, err
	}
	return res < 0, nil
}
```
This would match the logic in the semver PR  [here](https://github.com/optimizely/go-sdk/pull/289/files#diff-10bf306c78e43b7fe2b8736f4c50d92bR163-R197)

PR #291, opened 2020-08-25
fix: check for number in the semver prefix

thomaszurkan-optimizely:


pkg/decision/evaluator/matchers/semver.go
```diff
@@ -128,6 +128,12 @@ func (sv SemanticVersion) splitSemanticVersion(targetedVersion string) ([]string
 		return []string{}, errors.New(string(reasons.AttributeFormatInvalid))
 	}
 
+	for i := 0; i < len(targetedVersionParts);i++ {
+		if _, err := strconv.Atoi(targetedVersionParts[i]); err == nil {
```
msohailhussain, line 132: Why using Atoi, we may use IsNumber.

PR #292, opened 2020-08-28
Add support for sending flag decisions along with decision metadata.

mikecdavis:
## Summary
* Add experiment/feature flag key, variation key and type to impression events.
* Send events for **ALL** decision types.
* Add `Metadata` field to EventBatch.Decisions to capture flag type, key and variation key.

This change outline the proposed logical and structural changes required to support capturing all decision instances. The new eventbatch field is backwards compatible with the current pipeline.

### DIFF
--- a/pkg/event/events.go
+++ b/pkg/event/events.go
@@ -41,12 +41,19 @@ type UserEvent struct {
 // ImpressionEvent represents an impression event
 type ImpressionEvent struct {
 	Attributes   []VisitorAttribute
-	CampaignID   string            `json:"campaign_id"`
-	EntityID     string            `json:"entity_id"`
-	ExperimentID string            `json:"experiment_id"`
-	Key          string            `json:"key"`
-	Metadata     map[string]string `json:"metadata"`
-	VariationID  string            `json:"variation_id"`
+	CampaignID   string           `json:"campaign_id"`
+	EntityID     string           `json:"entity_id"`
+	ExperimentID string           `json:"experiment_id"`
+	Key          string           `json:"key"`
+	Metadata     DecisionMetadata `json:"metadata"`
+	VariationID  string           `json:"variation_id"`
+}
+
+// DecisionMetadata captures additional information regarding the decision
+type DecisionMetadata struct {
+	FlagType     string `json:"flag_type"`
+	FlagKey      string `json:"flag_key"`
+	VariationKey string `json:"variation_key"`
 }
 
 // ConversionEvent represents a conversion event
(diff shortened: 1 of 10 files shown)


mjc1283 (approved):

pkg/event/factory.go
```diff
@@ -57,26 +59,47 @@ func CreateEventContext(projectConfig config.ProjectConfig) Context {
...
+) ImpressionEvent {
+
+	metadata := map[string]string{
+		"FLAG_KEY":  flagKey,
+		"FLAG_TYPE": flagType,
+	}
+
+	var variationID string
+	if variation != nil {
+		metadata["VARIAITON_KEY"] = variation.Key
```
yasirfolio3, line 77: nit: should be VARIATION_KEY

pkg/event/factory.go
```diff
@@ -57,26 +59,47 @@ func CreateEventContext(projectConfig config.ProjectConfig) Context {
...
 
-	return impression
+	return event
 }
 
 // CreateImpressionUserEvent creates and returns ImpressionEvent for user
 func CreateImpressionUserEvent(projectConfig config.ProjectConfig, experiment entities.Experiment,
-	variation entities.Variation,
-	userContext entities.UserContext) UserEvent {
+	variation *entities.Variation, userContext entities.UserContext, flagKey, flagType string) (UserEvent, bool) {
```
msohailhussain, line 96: ٍ`variation *entities.Variation` won't it break existing apps? e.g. `Agent`?

mikecdavis: Unfortunately, this is public only due to our package structure and is technically should be a private function.

mikecdavis: FWIW, Agent does not call this directly

pkg/event/factory.go
mikecdavis: We've exposing the flaw behind open-ended maps :P

pkg/event/events.go
```diff
@@ -40,12 +40,13 @@ type UserEvent struct {
...
-	Key          string `json:"key"`
 	Attributes   []VisitorAttribute
-	VariationID  string `json:"variation_id"`
-	CampaignID   string `json:"campaign_id"`
-	ExperimentID string `json:"experiment_id"`
+	CampaignID   string            `json:"campaign_id"`
+	EntityID     string            `json:"entity_id"`
+	ExperimentID string            `json:"experiment_id"`
+	Key          string            `json:"key"`
+	Metadata     map[string]string `json:"metadata"`
```
msohailhussain, line 48: `strictly` typed representation would be great here.

pkg/event/factory.go
msohailhussain: got it.

pkg/event/events.go
mikecdavis: :+1: I've replaced this with a `struct`

msohailhussain (approved):
lgtm

pawels-optimizely (approved):
lgtm

pkg/event/events.go
```diff
@@ -40,12 +40,20 @@ type UserEvent struct {
...
+	Key          string           `json:"key"`
+	Metadata     DecisionMetadata `json:"metadata"`
+	VariationID  string           `json:"variation_id"`
+}
+
+// DecisionMetadata captures additional information regarding the decision
+type DecisionMetadata struct {
+	FlagType     string `json:"flag_type"`
+	FlagKey      string `json:"flag_key"`
+	VariationKey string `json:"variation_key"`
```
yasirfolio3, line 56: shouldn't this be defined with `omitempty` since this is an optional value?
```suggestion
	VariationKey string `json:"variation_key,omitempty"`
```

PR #297, opened 2020-10-19
feat(decide): add UserContext.

yasirfolio3:
## Summary
A part of multiple PRs for Decide API
- added OptimizelyUserContext
- added setUserContext API

## Tests
- Unit tests added.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -51,7 +51,7 @@ type OptimizelyClient struct {
 
 // CreateUserContext creates a context of the user for which decision APIs will be called.
 // A user context will be created successfully even when the SDK is not fully configured yet.
-func (o *OptimizelyClient) CreateUserContext(userContext entities.UserContext) *OptimizelyUserContext {
+func (o *OptimizelyClient) CreateUserContext(userContext entities.UserContext) OptimizelyUserContext {
 	return NewOptimizelyUserContext(o, userContext)
 }
 
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -26,6 +26,7 @@ import (
 	"strconv"
 
 	"github.com/optimizely/go-sdk/pkg/config"
+	"github.com/optimizely/go-sdk/pkg/decide"
 	"github.com/optimizely/go-sdk/pkg/decision"
 	"github.com/optimizely/go-sdk/pkg/entities"
 	"github.com/optimizely/go-sdk/pkg/event"
@@ -39,12 +40,36 @@ import (
 
 // OptimizelyClient is the entry point to the Optimizely SDK
 type OptimizelyClient struct {
-	ConfigManager      config.ProjectConfigManager
-	DecisionService    decision.Service
-	EventProcessor     event.Processor
-	notificationCenter notification.Center
-	execGroup          *utils.ExecGroup
-	logger             logging.OptimizelyLogProducer
+	ConfigManager        config.ProjectConfigManager
+	DecisionService      decision.Service
+	EventProcessor       event.Processor
+	notificationCenter   notification.Center
+	execGroup            *utils.ExecGroup
+	logger               logging.OptimizelyLogProducer
+	defaultDecideOptions []decide.Options
+}
+
+// CreateUserContext will return an OptimizelyUserContext associated with this OptimizelyClient
+// and context of the user for which decision APIs will be called
+//
+// The SDK will keep this context until it is called again with a different context data.
+//
+// - This API can be called after SDK initialization is completed (otherwise the __sdkNotReady__ error
+//	 will be returned).
+// - Only one user outstanding. The user-context can be changed any time by calling the same method
+//   with a different user-context value.
+// - The SDK will copy the parameter value to create an internal user-context data atomically,
+//   so any further change in its caller copy after the API call is not reflected into the SDK state.
+// - Once this API is called, the following other API calls can be called without a user-context
+//   parameter to use the same user-context.
+// - Each Decide API call can contain an optional user-context parameter when the call targets a
+//   different user-context. This optional user-context parameter value will be used once only,
+//   instead of replacing the saved user-context. This call-based context control can be used to
+//   support multiple users at the same time.
+// - If a user-context has not been set yet and decide APIs are called without a user-context parameter,
+//   SDK will return an error decision (__userNotSet__).
+func (o *OptimizelyClient) CreateUserContext(userContext entities.UserContext) *OptimizelyUserContext {
+	return NewOptimizelyUserContext(o, userContext)
 }
 
 // Activate returns the key of the variation the user is bucketed into and queues up an impression event to be sent to
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -49,25 +49,8 @@ type OptimizelyClient struct {
 	defaultDecideOptions []decide.Options
 }
 
-// CreateUserContext will return an OptimizelyUserContext associated with this OptimizelyClient
-// and context of the user for which decision APIs will be called
-//
-// The SDK will keep this context until it is called again with a different context data.
-//
-// - This API can be called after SDK initialization is completed (otherwise the __sdkNotReady__ error
-//	 will be returned).
-// - Only one user outstanding. The user-context can be changed any time by calling the same method
-//   with a different user-context value.
-// - The SDK will copy the parameter value to create an internal user-context data atomically,
-//   so any further change in its caller copy after the API call is not reflected into the SDK state.
-// - Once this API is called, the following other API calls can be called without a user-context
-//   parameter to use the same user-context.
-// - Each Decide API call can contain an optional user-context parameter when the call targets a
-//   different user-context. This optional user-context parameter value will be used once only,
-//   instead of replacing the saved user-context. This call-based context control can be used to
-//   support multiple users at the same time.
-// - If a user-context has not been set yet and decide APIs are called without a user-context parameter,
-//   SDK will return an error decision (__userNotSet__).
+// CreateUserContext creates a context of the user for which decision APIs will be called.
+// A user context will be created successfully even when the SDK is not fully configured yet.
 func (o *OptimizelyClient) CreateUserContext(userContext entities.UserContext) *OptimizelyUserContext {
 	return NewOptimizelyUserContext(o, userContext)
 }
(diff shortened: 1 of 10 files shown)


pkg/client/client_test.go
```diff
@@ -2447,6 +2447,38 @@ func (s *ClientTestSuiteFM) TestGetEnabledFeaturesErrorCases() {
 	s.mockDecisionService.AssertNotCalled(s.T(), "GetFeatureDecision")
 }
 
+func TestCreateUserContext(t *testing.T) {
```
msohailhussain, line 2450: call createusercontext, pass e.g. attributes1, 
modify any existing key, 
assert, getusercontext should not have that change.

pkg/client/factory_test.go
```diff
@@ -201,3 +202,17 @@ func TestClientWithDatafileAccessToken(t *testing.T) {
 
 	assert.Equal(t, accessToken, factory.DatafileAccessToken)
 }
+
+func TestClientWithDefaultDecideOptions(t *testing.T) {
```
msohailhussain, line 206: add same case, u modify decideOptions outside, that should not impact `optimizelyClient.defaultDecideOptions`

pkg/client/optimizely_decision.go
```diff
@@ -0,0 +1,95 @@
...
+		variables:    variables,
+		ruleKey:      ruleKey,
+		flagKey:      flagKey,
+		userContext:  userContext,
+		reasons:      reasons,
+	}
+}
+
+// CreateErrorDecision returns a decision with error
+func CreateErrorDecision(key string, user *OptimizelyUserContext, err error) OptimizelyDecision {
```
msohailhussain, line 49: `NewErrorDecision`

pkg/client/optimizely_decision.go
```diff
@@ -0,0 +1,95 @@
...
+	return o.userContext
+}
+
+// GetReasons returns an array of error/info/debug messages describing why the decision has been made.
+func (o OptimizelyDecision) GetReasons() []string {
+	return o.reasons
+}
+
+// HasFailed returns if variation was not found.
+func (o OptimizelyDecision) HasFailed() bool {
```
msohailhussain, line 93: what does it mean, will check later.

msohailhussain (changes requested):
have few comments

pkg/client/optimizely_decision.go
yasirfolio3: Removing it, will add it if required in the subsequent PR.

pkg/client/optimizely_user_context.go
```diff
@@ -0,0 +1,111 @@
...
+	"sync"
+
+	"github.com/optimizely/go-sdk/pkg/decide"
+	"github.com/optimizely/go-sdk/pkg/entities"
+)
+
+// OptimizelyUserContext defines user contexts that the SDK will use to make decisions for.
+type OptimizelyUserContext struct {
+	optimizely  *OptimizelyClient
+	userContext entities.UserContext
```
jaeopt, line 30: Can we replace this "userContext" with userId and attributes, then we do not need to mix UserContext and OptimizelyUserContext. Keep using UserContext for legacy APIs and use OptimizelyUserContext for decide APIs.

pkg/client/optimizely_decision.go
```diff
@@ -0,0 +1,90 @@
...
+		flagKey:      flagKey,
+		userContext:  userContext,
+		reasons:      reasons,
+	}
+}
+
+// NewErrorDecision returns a decision with error
+func NewErrorDecision(key string, user OptimizelyUserContext, err error) OptimizelyDecision {
+	return OptimizelyDecision{
+		flagKey:     key,
```
jaeopt, line 51: "variables" won't be null. Pass empty variables for error decision.

pkg/client/optimizely_decision_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (s *OptimizelyDecisionTestSuite) TestNewErrorDecision() {
+	flagKey := "flag1"
+	errorString := "SDK has an error"
+	userContext := entities.UserContext{ID: "testUser1", Attributes: map[string]interface{}{"key": 1212}}
+	optimizelyUserContext := s.OptimizelyClient.CreateUserContext(userContext)
+	decision := NewErrorDecision(flagKey, optimizelyUserContext, errors.New(errorString))
+
+	assert.Equal(s.T(), "", decision.GetVariationKey())
```
jaeopt, line 68: Uhm, we use null variationKey for decision error. GO staring cannot be null. Should be OK to check empty string for error checking.

jaeopt (changes requested):
A mixed use of UserContext and OptimizelyUserContext will be very confusing. We need to clean this up. I put some suggestions and let's discuss about the possible options.

pkg/decide/decide_options.go
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package decide has option definitions for decide api
+package decide
+
+// Options controlling flag decisions.
+type Options int
+
+const (
```
pawels-optimizely, line 23: I think it should be a struct:
```
type OptimizelyDecideOptions struct {
	DisableTracking bool
	// EnabledOnly when set, will return decisions for enabled-flags only.
	EnabledOnly bool
	// BypassUPS when set, will bypass UPS (both lookup and save) for decision
	BypassUPS bool
	// ForExperiment when set, will specify that the key parameter(s) of decide and decideAll APIs should be for experiments (not flags)
	ForExperiment bool
	// IncludeReasons when set, will return decision debugging messages in reasons
	IncludeReasons bool
}
```

otherwise it will be really awkward to check the condition.

pawels-optimizely (changes requested):
I think OptimizelyDecideOptions should be a struct

pkg/client/client.go
```diff
@@ -39,12 +40,19 @@ import (
...
+	EventProcessor       event.Processor
+	notificationCenter   notification.Center
+	execGroup            *utils.ExecGroup
+	logger               logging.OptimizelyLogProducer
+	defaultDecideOptions []decide.Options
+}
+
+// CreateUserContext creates a context of the user for which decision APIs will be called.
+// A user context will be created successfully even when the SDK is not fully configured yet.
+func (o *OptimizelyClient) CreateUserContext(userContext entities.UserContext) OptimizelyUserContext {
```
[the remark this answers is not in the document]
pawels-optimizely: we should have only one user context

pkg/client/factory.go
```diff
@@ -79,14 +79,12 @@ func (f *OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClie
...
-		logger:               logging.GetLogger(f.SDKKey, "OptimizelyClient"),
-		defaultDecideOptions: []decide.Options{},
+		notificationCenter: registry.GetNotificationCenter(f.SDKKey),
+		logger:             logging.GetLogger(f.SDKKey, "OptimizelyClient"),
 	}
 
-	if len(f.defaultDecideOptions) > 0 {
-		appClient.defaultDecideOptions = f.defaultDecideOptions
-	}
+	copiedOptions := f.defaultDecideOptions
```
pawels-optimizely, line 86: why are we doing this?

yasirfolio3: removing this since its not needed here, thanks for pointing out 👍

pkg/client/factory.go
```diff
@@ -78,7 +80,10 @@ func (f *OptimizelyFactory) Client(clientOptions ...OptionFunc) (*OptimizelyClie
 	eg := utils.NewExecGroup(ctx, logging.GetLogger(f.SDKKey, "ExecGroup"))
 	appClient := &OptimizelyClient{execGroup: eg,
 		notificationCenter: registry.GetNotificationCenter(f.SDKKey),
-		logger:             logging.GetLogger(f.SDKKey, "OptimizelyClient")}
+		logger:             logging.GetLogger(f.SDKKey, "OptimizelyClient"),
+	}
+
+	appClient.defaultDecideOptions = f.defaultDecideOptions
```
pawels-optimizely, line 86: could be moved to initial declaration

pkg/client/factory_test.go
```diff
@@ -204,7 +204,10 @@ func TestClientWithDatafileAccessToken(t *testing.T) {
 }
 
 func TestClientWithDefaultDecideOptions(t *testing.T) {
-	decideOptions := []decide.Options{decide.DisableDecisionEvent, decide.EnabledFlagsOnly}
+	decideOptions := decide.OptimizelyDecideOptions{
```
jaeopt, line 207: I discussed with @pawels-optimizely about this change from array to struct for options. The array of options still looks good to me :) I'd like to keep this open and see how the options will be implemented in the following PRs.

pkg/client/optimizely_user_context.go
```diff
@@ -26,18 +26,23 @@ import (
 
 // OptimizelyUserContext defines user contexts that the SDK will use to make decisions for.
 type OptimizelyUserContext struct {
-	optimizely  *OptimizelyClient
-	userContext entities.UserContext
-	mutex       *sync.RWMutex
+	entities.UserContext
```
jaeopt, line 29: Can we keep "ID" and "attributes" here instead of UserContext? In this way, decide-api users won't see the confusion about 2 user-contexts. We can also remove the dependency onto the legacy userContext and in the long term, will have a better chance to clean up the 2 context confusion.

jaeopt (changes requested):
If possible, I'd like to remove the legacy userContext from OptimizelyUserContext, so decide-api users won't see 2 user contexts. Any downside for keeping {ID, attribute} pair in here?

pkg/client/optimizely_user_context.go
```diff
@@ -0,0 +1,109 @@
...
+// OptimizelyUserContext defines user contexts that the SDK will use to make decisions for.
+type OptimizelyUserContext struct {
+	userID     string
+	attributes map[string]interface{}
+	optimizely *OptimizelyClient
+	mutex      *sync.RWMutex
+}
+
+// returns an instance of the optimizely user context.
+func newOptimizelyUserContext(optimizely *OptimizelyClient, userID string, attributes map[string]interface{}) OptimizelyUserContext {
```
pawels-optimizely, line 36: it could take old context instead of userId and attributes

pawels-optimizely (approved):
lgtm, one nit comment

jaeopt (approved):
LGTM

pkg/client/optimizely_user_context.go
jaeopt: I recommend not to support the old context here. It can be a bit less convenient but will be much better to keep it clean from old context.

msohailhussain (approved):
lgtm

PR #299, opened 2020-10-28
feat(decide): add decide, decideAll and decideForKeys API.

yasirfolio3:
## Summary
A part of multiple PRs for Decide API
- added decide, decideAll and decideForKeys API

## Tests
- Unit tests added.


pkg/client/client.go
```diff
@@ -57,6 +57,157 @@ func (o *OptimizelyClient) CreateUserContext(userID string, attributes map[strin
...
+			default:
+				err = errors.New("unexpected error")
+			}
+			errorMessage := fmt.Sprintf("decideForKeys call, optimizely SDK is panicking with the error:")
+			o.logger.Error(errorMessage, err)
+			o.logger.Debug(string(debug.Stack()))
+		}
+	}()
+
+	decisionMap := map[string]OptimizelyDecision{}
```
msohailhussain, line 157: bring it after all checks.

msohailhussain (changes requested):
please address.

pkg/client/client.go
```diff
@@ -57,6 +57,157 @@ func (o *OptimizelyClient) CreateUserContext(userID string, attributes map[strin
...
+		Attributes: userContext.GetUserAttributes(),
+	}
+	var variationKey string
+	eventSent := false
+	flagEnabled := false
+	allOptions := o.getAllOptions(options)
+	decisionReasons := decide.NewDecisionReasons(options)
+	decisionContext.Variable = entities.Variable{}
+
+	featureDecision, _ := o.DecisionService.GetFeatureDecision(decisionContext, usrContext, options, decisionReasons)
```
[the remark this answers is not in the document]
yasirfolio3: that is handled here: 
https://github.com/optimizely/go-sdk/pull/299/files#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R85

pkg/client/client.go
```diff
@@ -57,6 +57,157 @@ func (o *OptimizelyClient) CreateUserContext(userID string, attributes map[strin
...
+		}
+	}
+
+	variableMap := map[string]interface{}{}
+	if !allOptions.ExcludeVariables {
+		variableMap = o.getDecisionVariableMap(feature, featureDecision.Variation, flagEnabled, decisionReasons)
+	}
+	optimizelyJSON := optimizelyjson.NewOptimizelyJSONfromMap(variableMap)
+	reasonsToReport := decisionReasons.ToReport()
+	ruleKey := featureDecision.Experiment.Key
```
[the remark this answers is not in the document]
yasirfolio3: it will always be an object.

pkg/client/client.go
yasirfolio3: this is being used by two of the checks immediately after it.

pkg/client/client.go
```diff
@@ -57,6 +57,153 @@ func (o *OptimizelyClient) CreateUserContext(userID string, attributes map[strin
...
+		return NewErrorDecision(key, userContext, decide.GetDecideError(decide.FlagKeyInvalid, key))
+	}
+	decisionContext.Feature = &feature
+
+	usrContext := entities.UserContext{
+		ID:         userContext.GetUserID(),
+		Attributes: userContext.GetUserAttributes(),
+	}
+	var variationKey string
+	eventSent := false
```
pawels-optimizely, line 96: var eventSent, flagEnabled bool

pkg/client/client.go
```diff
@@ -57,6 +57,153 @@ func (o *OptimizelyClient) CreateUserContext(userID string, attributes map[strin
...
+	flagEnabled := false
+	allOptions := o.getAllOptions(options)
+	decisionReasons := decide.NewDecisionReasons(options)
+	decisionContext.Variable = entities.Variable{}
+
+	featureDecision, _ := o.DecisionService.GetFeatureDecision(decisionContext, usrContext, options, decisionReasons)
```
pawels-optimizely, line 102: never leave err in the production code. at least a warning has to be flagged.

pkg/client/optimizely_user_context.go
```diff
@@ -46,17 +49,17 @@ func newOptimizelyUserContext(optimizely *OptimizelyClient, userID string, attri
...
 
 // GetUserID returns userID for Optimizely user context
-func (o *OptimizelyUserContext) GetUserID() string {
+func (o OptimizelyUserContext) GetUserID() string {
 	return o.userID
 }
 
 // GetUserAttributes returns user attributes for Optimizely user context
-func (o *OptimizelyUserContext) GetUserAttributes() map[string]interface{} {
+func (o OptimizelyUserContext) GetUserAttributes() map[string]interface{} {
```
pawels-optimizely, line 62: why do we need a copy of user attributes ? are we allowed to modify them?

pawels-optimizely (commented):
overall, looks ok, a few nit comments.

Also, I would try to pass bigger objects by reference, like  decide.OptimizelyDecideOptions

pkg/decision/flag_notification.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // FlagNotification constructs default flag notification
-func FlagNotification(flagKey, variationKey, ruleKey string, enabled, decisionEventDispatched bool, userContext *entities.UserContext, variables map[string]interface{}, reasons []string) *notification.DecisionNotification {
```
pawels-optimizely, line 26: what was the reason for removing the passing by the pointer of user context ?

pkg/client/optimizely_user_context_test.go
```diff
@@ -119,6 +137,574 @@ func (s *OptimizelyUserContextTestSuite) TestSetAttributeOverride() {
...
+	decision := userContext.Decide(flagKey, decide.OptimizelyDecideOptions{})
+	// should return variationId2 set by UPS
+	s.Equal(variationKey2, decision.GetVariationKey())
+
+	decision = userContext.Decide(flagKey, decide.OptimizelyDecideOptions{IgnoreUserProfileService: true})
+	// should ignore variationId2 set by UPS and return variationId1
+	s.Equal(variationKey1, decision.GetVariationKey())
+	// also should not save either
+	userProfileService.AssertNotCalled(s.T(), "Save", mock.Anything)
+}
```
jaeopt, line 582: Need a test validating IgnoreUPS option bypasses UPS save as well.

pkg/client/optimizely_user_context_test.go
```diff
@@ -119,6 +137,574 @@ func (s *OptimizelyUserContextTestSuite) TestSetAttributeOverride() {
...
+	decision = user.Decide(flagKey, options)
+	s.Len(decision.GetReasons(), 1)
+	s.Equal(decide.GetDecideMessage(decide.FlagKeyInvalid, flagKey), decision.GetReasons()[0])
+
+	// valid flag key
+	options.IncludeReasons = false
+	flagKey = "feature_1"
+	decision = user.Decide(flagKey, options)
+	s.Len(decision.GetReasons(), 0)
+}
```
jaeopt, line 618: Need one more case: IncludeReasons = true and flagKey = "feature_1". Then reasons.count > 0

pkg/client/optimizely_user_context_test.go
```diff
@@ -119,6 +137,574 @@ func (s *OptimizelyUserContextTestSuite) TestSetAttributeOverride() {
...
+func (s *OptimizelyUserContextTestSuite) TestDefaultDecideOptions() {
+	flagKey := "feature_1"
+	options := decide.OptimizelyDecideOptions{ExcludeVariables: true}
+	client, _ := s.factory.Client(WithEventProcessor(s.eventProcessor), WithDefaultDecideOptions(options))
+	userContext := client.CreateUserContext(s.userID, nil)
+
+	// should be excluded by DefaultDecideOption
+	decision := userContext.Decide(flagKey, decide.OptimizelyDecideOptions{})
+	s.Len(decision.GetVariables().ToMap(), 0)
+}
```
jaeopt, line 629: Can we add one more test with DefaultDecideOptions with "EnabledFlagsOnly". It takes a different path from other options. Worth to have the test.

jaeopt: Also need some tests validating DefaultDecideOptions and api parameter options are merged and work together ok.

jaeopt (changes requested):
It looks good! 
More test cases suggested. Also I see all convenience methods without options are missing.

pkg/client/optimizely_user_context.go
yasirfolio3: The reason for returning a copy of the attributes is to not allow user to make changes to the stored attributes directly by calling `GetUserAttributes` and making changes to it since `map` is a pointer type. To test this behaviour, a test can be seen here:
https://github.com/optimizely/go-sdk/pull/299/files#diff-97b3b48891edc303a7563bdf3ec2adfc96cc4819bdca644bc277d29a78d55d2cR75

pkg/decision/flag_notification.go
yasirfolio3: This was done so that the following methods can be called from the `OptimizelyUserContext` returned in the `decision` api response, Since we do not return a pointer of `OptimizelyUserContext`.

1. `GetOptimizely`
2. `GetUserID` 
3. `GetUserAttributes `

pkg/client/client.go
```diff
@@ -57,6 +57,153 @@ func (o *OptimizelyClient) CreateUserContext(userID string, attributes map[strin
...
+	if _, err = o.getProjectConfig(); err != nil {
+		o.logger.Error("Optimizely instance is not valid, failing decideForKeys call.", err)
+		return decisionMap
+	}
+
+	if len(keys) == 0 {
+		return decisionMap
+	}
+
+	enabledFlagsOnly := o.getAllOptions(options).EnabledFlagsOnly
```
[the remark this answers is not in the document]
yasirfolio3: Since `decideForKeys` calls `decide` internally and `decide` itself calls `allOptions` method, we are only using `allOptions` here to check `EnabledFlagsOnly` bool.

pkg/client/optimizely_user_context.go
```diff
@@ -75,17 +78,17 @@ func (o *OptimizelyUserContext) SetAttribute(key string, value interface{}) {
 // Decide returns a decision result for a given flag key and a user context, which contains
 // all data required to deliver the flag or experiment.
 func (o *OptimizelyUserContext) Decide(key string, options decide.OptimizelyDecideOptions) OptimizelyDecision {
```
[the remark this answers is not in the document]
yasirfolio3: method `overloading` is not possible in go, instead we will have to create methods with new names. @msohailhussain

pkg/client/optimizely_user_context_test.go
```diff
@@ -119,6 +137,574 @@ func (s *OptimizelyUserContextTestSuite) TestSetAttributeOverride() {
...
+
+func (s *OptimizelyUserContextTestSuite) TestDecideNullVariation() {
+	flagKey := "feature_3"
+	variablesExpected := optimizelyjson.NewOptimizelyJSONfromMap(map[string]interface{}{})
+
+	var options decide.OptimizelyDecideOptions
+	user := s.OptimizelyClient.CreateUserContext(s.userID, nil)
+	decision := user.Decide(flagKey, options)
+
+	s.Equal("", decision.GetVariationKey())
```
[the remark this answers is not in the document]
yasirfolio3: in go, string values cannot be nil and always default to empty string.

pkg/client/optimizely_user_context_test.go
yasirfolio3: Implementation of reasons will be added in a separate PR, adding `to do` for now.

pkg/client/optimizely_user_context.go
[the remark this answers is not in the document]
jaeopt: I see. As long as this is consistent with existing APIs, it should be ok.

jaeopt (approved):
Looks good. A couple of name changes suggested.

pkg/client/client.go
```diff
@@ -46,7 +46,7 @@ type OptimizelyClient struct {
 	notificationCenter   notification.Center
 	execGroup            *utils.ExecGroup
 	logger               logging.OptimizelyLogProducer
-	defaultDecideOptions decide.OptimizelyDecideOptions
+	defaultDecideoptions *decide.OptimizelyDecideOptions
```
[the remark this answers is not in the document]
yasirfolio3: nit fixed. Yes it's private, we take arguments as an array from the user but keep it as a struct internally for ease of access and not having to iterate through the array every time.

jaeopt (approved):
LGTM

pkg/decision/flag_notification.go
yasirfolio3: Sorry, this was meant for some other comment, please discard 👍

pawels-optimizely (approved):

msohailhussain (approved):
lgtm

PR #301, opened 2020-11-09
feat: added "enabled" field to decision metadata structure

pawels-optimizely:
## Summary
- added "enabled" field to decision metadata structure
- setting "enabled" to true for activate() since we check for non nil variation.


msohailhussain (approved):
lgtm

mjc1283 (approved):

PR #303, opened 2020-12-22
refact(decision-reasons): Refactoring reasons for decide api

yasirfolio3:
## Summary
Reasons were previously being collected by sending a reasons object down the stack to all the related functions which would mutate it by pushing their own reasons. This is now changed. Every related function now returns the reasons array along with the actual value.

## Test plan
- Manually tested thoroughly.
- Fixed all the affected unit test.
- FSC passing

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -96,7 +96,9 @@ func (o *OptimizelyClient) decide(userContext OptimizelyUserContext, key string,
 	decisionReasons := decide.NewDecisionReasons(&allOptions)
 	decisionContext.Variable = entities.Variable{}
 
-	featureDecision, err := o.DecisionService.GetFeatureDecision(decisionContext, usrContext, &allOptions, decisionReasons)
+	featureDecision, reasons, err := o.DecisionService.GetFeatureDecision(decisionContext, usrContext, &allOptions)
+	decisionReasons.Append(reasons)
+
 	if err != nil {
 		o.logger.Warning(fmt.Sprintf(`Received error while making a decision for feature "%s": %s`, key, err))
 	}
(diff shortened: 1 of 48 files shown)


pkg/client/client.go
```diff
@@ -808,7 +811,7 @@ func (o *OptimizelyClient) getFeatureDecision(featureKey, variableKey string, us
 
 	decisionContext.Variable = variable
 	options := &decide.Options{}
-	featureDecision, err = o.DecisionService.GetFeatureDecision(decisionContext, userContext, options, decide.NewDecisionReasons(options))
+	featureDecision, _, err = o.DecisionService.GetFeatureDecision(decisionContext, userContext, options)
```
msohailhussain, line 814: why `_`, I would suggest, if reason is returned, add it regarding anything is added in reasons or not.

pkg/client/client.go
```diff
@@ -839,7 +842,7 @@ func (o *OptimizelyClient) getExperimentDecision(experimentKey string, userConte
 	}
 
 	options := &decide.Options{}
-	experimentDecision, err = o.DecisionService.GetExperimentDecision(decisionContext, userContext, options, decide.NewDecisionReasons(options))
+	experimentDecision, _, err = o.DecisionService.GetExperimentDecision(decisionContext, userContext, options)
```
msohailhussain, line 845: same across all code changes.

pkg/client/client.go
pawels-optimizely: I agree, but that function does not do anything with reasons - it does not return them

pkg/client/client.go
pawels-optimizely: yeah, but this function does not return reasons

pawels-optimizely (approved):
lgtm, some functions return empty (default) reasons, but we can leave it just in case we want to populate them later on.

msohailhussain (approved):
lgtm.

PR #304, opened 2020-12-22
refact(decide-reasons): Adding new info and error messages to decide reasons.

yasirfolio3:
## Summary
Added new reasons for decide api.

## Test plan
- Added and updated unit tests.
- FSC passing.

### DIFF
--- a/pkg/decision/evaluator/matchers/exact.go
+++ b/pkg/decision/evaluator/matchers/exact.go
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2019-2020, Optimizely, Inc. and contributors                   *
+ * Copyright 2019-2021, Optimizely, Inc. and contributors                   *
  *                                                                          *
  * Licensed under the Apache License, Version 2.0 (the "License");          *
  * you may not use this file except in compliance with the License.         *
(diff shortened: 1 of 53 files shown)


pkg/decision/evaluator/matchers/exact.go
```diff
@@ -27,41 +27,42 @@ import (
 )
 
 // ExactMatcher matches against the "exact" match type
-func ExactMatcher(condition entities.Condition, user entities.UserContext, logger logging.OptimizelyLogProducer, reasons decide.DecisionReasons) (bool, error) {
+func ExactMatcher(condition entities.Condition, user entities.UserContext, logger logging.OptimizelyLogProducer) (bool, decide.DecisionReasons, error) {
+	reasons := decide.NewDecisionReasons(nil)
```
jaeopt, line 31: Looks like we do not use "reasons" in most matchers and conditions, just adding a lot of boilerplate codes. Can we remove "reasons" from them.
I understand one case above we have meaning messages. We can consider to use "err" message (or if it's not easy, we can just drop from reasons - they can use debug message why this happened).

jaeopt (changes requested):
It's a lot of work and overall looks great! 
A few changes suggested. I also like to add a few more tests for validating core reasons messages.

jaeopt (changes requested):
All changes look good! One more small change suggested.

jaeopt (approved):
LGTM

PR #329, opened 2021-12-09
fix(utils): logging correct error message in HTTP requests

seanpfeifer:
This is currently set to always log a `nil` error using the unset return parameter `e`, instead of the actual error that gets assigned to `doErr`.

This change will allow users to see the HTTP error in these messages.

## Submissions info
* CLA has been signed
* Test has been added and is passing
  * Confirmed that the test fails in the event of a regression


seanpfeifer:
The Travis CI `Integration Tests` step has failed with the following:
```sh
$ mkdir $HOME/travisci-tools && pushd $HOME/travisci-tools && git init && git pull https://$CI_USER_TOKEN@github.com/optimizely/travisci-tools.git && popd
~/travisci-tools ~/build/optimizely/go-sdk
hint: ... (omitted for log brevity)
Initialized empty Git repository in /home/travis/travisci-tools/.git/
remote: Repository not found.
fatal: Authentication failed for 'https://github.com/optimizely/travisci-tools.git/'
The command "mkdir $HOME/travisci-tools && pushd $HOME/travisci-tools && git init && git pull https://$CI_USER_TOKEN@github.com/optimizely/travisci-tools.git && popd" failed and exited with 1 during .
```

From these logs it looks like something may be missing / misconfigured with Travis or the `optimizely/travisci-tools` repo.

pulak-opti:
Hi @seanpfeifer, thanks for opening this PR!

Can you please update your branch with master?

seanpfeifer:
Merged and tested.

pkg/utils/requester_test.go
```diff
@@ -190,12 +190,36 @@ func TestPostObj(t *testing.T) {
...
+// mockLogger exists to check that the logged error is what we expected
+type mockLogger struct {
+	Errors []error
+}
+
+func (t *mockLogger) Debug(message string)   {}
+func (t *mockLogger) Info(message string)    {}
+func (t *mockLogger) Warning(message string) {}
+func (t *mockLogger) Error(message string, err interface{}) {
+	t.Errors = append(t.Errors, err.(error))
```
pulak-opti, line 202: It can cause panic in two cases:
1. when err interface{} is nil 
2. when err interface{} is not error type

We can avoid those by following way:
`
if _, ok := err.(error); ok {
      t.Errors = append(t.Errors, err)
}
`
What do you think?

pkg/utils/requester_test.go
```diff
@@ -190,12 +190,36 @@ func TestPostObj(t *testing.T) {
...
 
-	httpreq := NewHTTPRequester(logging.GetLogger("", ""))
-	_, _, _, err := httpreq.Get("blah12345/good")
-	_, ok := err.(*url.Error)
+	badURL := "blah12345/good"
+	_, _, _, err := httpreq.Get(badURL)
+	returnedErr, ok := err.(*url.Error)
 	assert.True(t, ok, "url error")
+
+	// Did we "log" the error?
```
pulak-opti, line 215:
```suggestion
	// Check to make sure we have some log for bad url
```

pulak-opti (changes requested):

pkg/utils/requester_test.go
seanpfeifer: Sounds good - added this check to `Error()`.

Also added an early failure in the dependent test function in the case where we don't have the expected error.

pkg/utils/requester_test.go
seanpfeifer: Changed comment to suggestion

yasirfolio3 (approved):
LGTM, can you please update the headers with the current year `2023` in both files.

seanpfeifer:
> LGTM, can you please update the headers with the current year `2023` in both files.

Updated!

pulak-opti:
Hi, the fix is merged here. https://github.com/optimizely/go-sdk/pull/374

Thank you for your contribution!

PR #339, opened 2022-08-23
OASIS-8551 fix(attribute-validation): Adds validation on user attributes to only allow primitive types.

yasirfolio3:
## Summary
- Adds validation on user attributes to only allow primitive types.
- Adds unit tests.

## Test plan
All tests should pass

# ISSUE
OASIS-8551

### DIFF
--- a/pkg/event/factory.go
+++ b/pkg/event/factory.go
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2019-2020, Optimizely, Inc. and contributors                   *
+ * Copyright 2019-2020,2022 Optimizely, Inc. and contributors               *
  *                                                                          *
  * Licensed under the Apache License, Version 2.0 (the "License");          *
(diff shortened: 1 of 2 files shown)


Mat001 (approved):
Looks good to me. 
The valid types are those that are supported by the data backend, make sure that is aligned.

jaeopt (approved):
LGTM

pkg/event/factory.go
```diff
@@ -291,3 +291,23 @@ func getTagValue(eventTags map[string]interface{}) (float64, error) {
 
 	return 0, errors.New("no event tag found for value")
 }
+
+// check if attribute value is valid
+func isValidAttribute(value interface{}) bool {
```
msohailhussain, line 296: can be in switch case.

msohailhussain (approved):
lgtm

PR #341, opened 2022-08-31
nit(attribute-validation): Using switch instead of multiple if checks.

yasirfolio3:
## Fixes:
- Using switch instead of multiple if checks.
- Covered all go basic types defined here: https://go.dev/tour/basics/11

## Tests:
- All tests should pass

## Jira ticket:
https://optimizely.atlassian.net/browse/OASIS-8593


pkg/event/factory.go
```diff
@@ -297,17 +297,10 @@ func isValidAttribute(value interface{}) bool {
...
-	}
-	if _, ok := value.(float64); ok {
-		return true
-	}
-	if _, ok := value.(int); ok {
-		return true
-	}
-	if _, ok := value.(bool); ok {
+	switch value.(type) {
+	case string, float64, int, bool:
```
jaeopt, line 301: Does this change work for float32, int8, which should be accepted?

jaeopt (changes requested):
A  clarification

pkg/event/factory.go
msohailhussain: https://go.dev/tour/basics/11
should give all types?

yasirfolio3: have covered all basic types now, please check 👍

jaeopt (approved):
LGTM
wondering what's wrong with the original if-based one. It'll cover all types.

yasirfolio3:
> LGTM wondering what's wrong with the original if-based one. It'll cover all types.

i tried the new tests for both new and previous code, behaviour was the same. We would've required `if` checks for all types too.

jaeopt (approved):
LGTM

msohailhussain (approved):
lgtm

PR #345, opened 2022-09-30
fix(visitor-timestamp): Fixes issue with visitor timestamp .

yasirfolio3:
## Summary
- Fixes issue with visitor where dispatch timestamp was used instead of userEvent creation timestamp.

## Issues
- FSSDK-8641

## Tests
- Unit and integration tests should pass.


jaeopt (approved):
LGTM

msohailhussain (approved):
lgtm

PR #347, opened 2022-10-05
feat: Added a generic LRUCache interface and a default implementation

yasirfolio3:
## Summary
- Added an interface for LRU Caching mechanism.  Include a default implementation.
- Generics were not used since they were introduced in go 1.18 whereas we provide support up-to go 1.12.

## Test plan
Included new tests in the test project

## Issues
FSSDK-8513

### DIFF
--- a/pkg/odp/lru_cache.go
+++ b/pkg/odp/lru_cache.go
@@ -23,6 +23,13 @@ import (
 	"time"
 )
 
+// Cache is used for caching ODP segments
+type Cache interface {
+	Save(key string, value interface{})
+	Lookup(key string) interface{}
+	Reset()
+}
+
 type cacheElement struct {
 	data   interface{}
 	time   int64
--- a/pkg/odp/lru_cache.go
+++ b/pkg/odp/lru_cache.go
@@ -1,0 +1,103 @@
+/****************************************************************************
+ * Copyright 2022, Optimizely, Inc. and contributors                        *
+ *                                                                          *
+ * Licensed under the Apache License, Version 2.0 (the "License");          *
+ * you may not use this file except in compliance with the License.         *
+ * You may obtain a copy of the License at                                  *
+ *                                                                          *
+ *    http://www.apache.org/licenses/LICENSE-2.0                            *
+ *                                                                          *
+ * Unless required by applicable law or agreed to in writing, software      *
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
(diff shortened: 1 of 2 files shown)


jaeopt (approved):
Looks great - with a clarification

pkg/odp/lru_cache.go
```diff
@@ -0,0 +1,110 @@
...
+type LRUCache struct {
+	queue         *list.List
+	items         map[string]*cacheElement
+	maxSize       int
+	timeoutInSecs int64
+	lock          *sync.RWMutex
+}
+
+// NewLRUCache returns a new instance of Least Recently Used in-memory cache
+func NewLRUCache(size int, timeoutInSecs int64) LRUCache {
```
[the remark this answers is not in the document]
yasirfolio3: To avoid converting `timeoutInSecs` to in64 every time `isValid` is called since `time.Now().Unix()` returns a `int64` value and calculations are only allowed between values of similar type's.

pkg/odp/lru_cache.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// Cache is used for caching ODP segments
+type Cache interface {
+	Save(key string, value interface{})
+	Lookup(key string) interface{}
+	Reset()
+}
+
+type cacheElement struct {
```
msohailhussain, line 33: `Cache` ?

pkg/odp/lru_cache.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"container/list"
+	"sync"
+	"time"
+)
+
+// Cache is used for caching ODP segments
+type Cache interface {
+	Save(key string, value interface{})
```
msohailhussain, line 28: can we add specific type of `interface` instead of `interface{}`

pkg/odp/lru_cache.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Save stores a new element into the cache
+func (l *LRUCache) Save(key string, value interface{}) {
+	if l.maxSize <= 0 {
+		return
+	}
+	l.lock.Lock()
+	defer l.lock.Unlock()
+	if item, ok := l.items[key]; !ok {
+		if l.maxSize == len(l.items) {
```
msohailhussain, line 61: add comment. what this logic does.

msohailhussain (approved):
please address my comments and lgtm!!!

pkg/odp/lru_cache.go
yasirfolio3: interface has no further types in go.

PR #355, opened 2022-11-24
feat(ats): add ODP integration.

yasirfolio3:
## Summary
This PR adds a support for Optimizely Data Platform (ODP) integration to Full Stack. With this extension, clients may not need to pre-determine and include user segments in attributes. SDK can fetch user segments from the ODP server for the current user. 

- Add a new public API to OptimizelyUserContext (fetchQualifiedSegments).
- Add a new public API to OptimizelyClient (SendOdpEvent).

## Test plan
- Tests for OptimizelyUserContext new APIs.
- Tests for OptimizelyClient new APIs.
- Tests for odpManager.

## Issues
- FSSDK-8516


pkg/client/client.go
```diff
@@ -54,6 +59,10 @@ type OptimizelyClient struct {
 // A user context will be created successfully even when the SDK is not fully configured yet.
 func (o *OptimizelyClient) CreateUserContext(userID string, attributes map[string]interface{}) OptimizelyUserContext {
 	// Passing qualified segments as nil initially since they will be fetched later
+	if o.identify {
```
msohailhussain, line 62: is the comment above is valid for this line?

pkg/client/client.go
```diff
@@ -234,6 +243,80 @@ func (o *OptimizelyClient) decideAll(userContext OptimizelyUserContext, options
 	return o.decideForKeys(userContext, allFlagKeys, options)
 }
 
+func (o *OptimizelyClient) fetchQualifiedSegments(userContext *OptimizelyUserContext, options []pkgOdpSegment.OptimizelySegmentOption, callback func(segments []string, err error)) {
```
msohailhussain, line 246: documentation?

pkg/client/client.go
```diff
@@ -234,6 +243,80 @@ func (o *OptimizelyClient) decideAll(userContext OptimizelyUserContext, options
...
+			o.logger.Error("fetchQualifiedSegments call, optimizely SDK is panicking with the error:", err)
+			o.logger.Debug(string(debug.Stack()))
+		}
+	}()
+
+	// on failure, qualifiedSegments should be reset if a previous value exists.
+	userContext.SetQualifiedSegments(nil)
+
+	_, err = o.getProjectConfig()
+	if err != nil {
```
msohailhussain, line 267: can't we combine these two conditions above & below.

pkg/client/client.go
```diff
@@ -234,6 +243,80 @@ func (o *OptimizelyClient) decideAll(userContext OptimizelyUserContext, options
...
+		if segmentsError == nil {
+			userContext.SetQualifiedSegments(qualifiedSegments)
+		}
+		return qualifiedSegments, segmentsError
+	}
+
+	// Async
+	if callback != nil {
+		go func() {
+			callback(fetchAndSetSegments())
```
msohailhussain, line 285: there may be inconsistency / race condition issue when you call routine. thoughts?

pkg/client/factory.go
```diff
@@ -38,15 +42,18 @@ type OptimizelyFactory struct {
...
-	configManager        config.ProjectConfigManager
-	ctx                  context.Context
-	decisionService      decision.Service
-	defaultDecideOptions *decide.Options
-	eventDispatcher      event.Dispatcher
-	eventProcessor       event.Processor
-	userProfileService   decision.UserProfileService
-	overrideStore        decision.ExperimentOverrideStore
-	metricsRegistry      metrics.Registry
+	identify              *bool
```
msohailhussain, line 45: alphabetize please.

pkg/client/factory.go
```diff
@@ -253,6 +286,81 @@ func (f *OptimizelyFactory) StaticClient() (optlyClient *OptimizelyClient, err e
 	return optlyClient, err
 }
 
+func (f *OptimizelyFactory) initializeOdpManager(appClient *OptimizelyClient) {
```
msohailhussain, line 289: will review this method later

pkg/config/datafileprojectconfig/config.go
```diff
@@ -273,7 +273,12 @@ func NewDatafileProjectConfig(jsonDatafile []byte, logger logging.OptimizelyLogP
 
 	var hostForODP, publicKeyForODP string
 	for _, integration := range datafile.Integrations {
-		if integration.Key == "odp" {
+		if integration.Key == nil {
```
msohailhussain, line 276: What about if it's empty.

pkg/config/datafileprojectconfig/entities/entities.go
```diff
@@ -106,9 +106,9 @@ type Rollout struct {
 
 // Integration represents a integration from the Optimizely datafile
 type Integration struct {
-	Key       string `json:"key"`
-	Host      string `json:"host"`
-	PublicKey string `json:"publicKey"`
+	Key       *string `json:"key"`
```
msohailhussain, line 109: take it as `type`

msohailhussain (changes requested):
few minor comments, will review again.

pkg/client/client.go
yasirfolio3: All internal implementations are using locks where required so this should not be a problem.

pkg/config/datafileprojectconfig/config.go
yasirfolio3: empty key is allowed.

pkg/config/datafileprojectconfig/entities/entities.go
yasirfolio3: do you mean we should rename it to type? pointer to the `string` is required here since `string` type in golang cannot be nil by itself.

pkg/client/factory.go
```diff
@@ -253,6 +286,81 @@ func (f *OptimizelyFactory) StaticClient() (optlyClient *OptimizelyClient, err e
...
+		identify = *f.identify
+	}
+	f.identify = &identify
+	appClient.identify = identify
+
+	if f.optimizelySDKSettings == nil {
+		f.optimizelySDKSettings = &OptimizelySdkSettings{
+			SegmentsCacheSize:          pkgOdpUtils.DefaultSegmentsCacheSize,
+			SegmentsCacheTimeoutInSecs: pkgOdpUtils.DefaultSegmentsCacheTimeout,
+		}
```
jaeopt, line 302: What is this structuring for? I see `optimizelySDKSettings` is used in this function only.

pkg/client/factory.go
```diff
@@ -165,6 +177,27 @@ func WithPollingConfigManagerDatafileAccessToken(pollingInterval time.Duration,
...
+		f.optimizelySDKSettings = optimizelySdkSettings
+	}
+}
+
+// WithOdpManager sets odp manager on a client.
+func WithOdpManager(odpManager odp.Manager) OptionFunc {
+	return func(f *OptimizelyFactory) {
+		f.odpManager = odpManager
+	}
+}
```
jaeopt, line 192: It looks like we use combinations of config files (cacheSize, cacheTimeout, enabled) and builder method for ODP configuration. Can we consider removing config files and adding them ODP Builder (withSegmentCacheSize, withSegmentCacheTimeout) just like other ODP configuration?

pkg/client/factory.go
```diff
@@ -253,6 +286,81 @@ func (f *OptimizelyFactory) StaticClient() (optlyClient *OptimizelyClient, err e
...
+
+	// Update user given odp manager with latest config
+	if f.odpManager != nil {
+		appClient.OdpManager = f.odpManager
+		// Update odp config with latest config
+		if isProjectConfigAvailable {
+			appClient.OdpManager.Update(projectConfig.GetPublicKeyForODP(), projectConfig.GetHostForODP(), projectConfig.GetSegmentList())
+		}
+	}
+}
```
jaeopt, line 329: Can we drop OMOption for OdpConfig?

```
	if odpManager == nil {
		odpManager = odp.NewOdpManager(f.SDKKey, ...)
	}

	if isProjectConfigAvailable {
		appClient.OdpManager.Update(projectConfig.GetPublicKeyForODP(), ...)
	}
```

pkg/odp/odp_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	SegmentManager segment.Manager
+	EventManager   event.Manager
+}
+
+// WithOdpConfig sets odpConfig option to be passed into the NewOdpManager method
+func WithOdpConfig(odpConfig config.Config) OMOptionFunc {
+	return func(om *DefaultOdpManager) {
+		om.OdpConfig = odpConfig
+	}
+}
```
jaeopt, line 57: We may drop this?

pkg/odp/odp_manager.go
```diff
@@ -0,0 +1,170 @@
...
+
+// WithSegmentsCache sets cache option to be passed into the NewOdpManager method
+func WithSegmentsCache(segmentsCache cache.Cache) OMOptionFunc {
+	return func(om *DefaultOdpManager) {
+		om.segmentsCache = segmentsCache
+	}
+}
+
```
jaeopt, line 65: We can add `withSegmentCacheSize` and `withSegmentCacheTimeout` here.

jaeopt (changes requested):
It looks good overall. It's a quite big PR. A few top-level suggestions first.

pkg/client/factory.go
```diff
@@ -253,6 +286,81 @@ func (f *OptimizelyFactory) StaticClient() (optlyClient *OptimizelyClient, err e
...
+	if f.optimizelySDKSettings == nil {
+		f.optimizelySDKSettings = &OptimizelySdkSettings{
+			SegmentsCacheSize:          pkgOdpUtils.DefaultSegmentsCacheSize,
+			SegmentsCacheTimeoutInSecs: pkgOdpUtils.DefaultSegmentsCacheTimeout,
+		}
+	}
+
+	projectConfig, err := appClient.ConfigManager.GetConfig()
+	// For cases when project config is not fetched yet
+	isProjectConfigAvailable := err == nil && projectConfig != nil
```
msohailhussain, line 307: I think return in case projectConfig not available.

pkg/config/datafileprojectconfig/entities/entities.go
msohailhussain: discussed offline.

msohailhussain (approved):
lgtm. please address jae's comments.

pkg/config/datafileprojectconfig/entities/entities.go
```diff
@@ -106,9 +106,9 @@ type Rollout struct {
...
+	Host      string  `json:"host"`
+	PublicKey string  `json:"publicKey"`
```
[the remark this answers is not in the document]
yasirfolio3: This is because in golang `string` cannot be null, it is always empty if nothing is assigned to it. In our case we need to make sure that `Key` is provided with every integration in the datafile, if not then client should throw an error. This is a internal struct and client will not be able to access it.

pkg/client/client.go
```diff
@@ -234,6 +243,81 @@ func (o *OptimizelyClient) decideAll(userContext OptimizelyUserContext, options
...
+			}
+			o.logger.Error("fetchQualifiedSegments call, optimizely SDK is panicking with the error:", err)
+			o.logger.Debug(string(debug.Stack()))
+		}
+	}()
+
+	// on failure, qualifiedSegments should be reset if a previous value exists.
+	userContext.SetQualifiedSegments(nil)
+
+	if _, err = o.getProjectConfig(); err != nil {
```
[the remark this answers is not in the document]
yasirfolio3: In case user provides an sdkKey, go-sdk allows user to create client as well as userContext while datafile is being downloaded in the background. As for discarding events, the check to ignore events if odp is not integrated is there in the odp manager.

jaeopt (approved):
LGTM with a nit.

pkg/client/optimizely_user_context.go
```diff
@@ -98,8 +98,16 @@ func (o *OptimizelyUserContext) SetAttribute(key string, value interface{}) {
...
-	o.optimizely.fetchQualifiedSegments(o, options, callback)
+func (o *OptimizelyUserContext) FetchQualifiedSegments(options []pkgOdpSegment.OptimizelySegmentOption) (success bool) {
+	o.optimizely.fetchQualifiedSegments(o, options, func(result bool) {
+		success = result
+	})
+	return
+}
+
+// FetchQualifiedSegmentsAsync fetches all qualified segments aysnchronously for the user context.
+func (o *OptimizelyUserContext) FetchQualifiedSegmentsAsync(callback func(success bool), options []pkgOdpSegment.OptimizelySegmentOption) {
```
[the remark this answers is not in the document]
yasirfolio3: it's not, might've copied it while looking into java, thank's for pointing out 👍

PR #356, opened 2023-01-09
refact(odp): Removes support for odp batchSize and fixes ticker.

yasirfolio3:
## Summary
- Removes support for changing odp batchSize. 
- If flush interval is set to 0, flushEvent timer is not started and batchSize is set to 1 to dispatch events immediately .
- Fixes a bug which restricted flushEvent timer to only send events using the odpConfiguration it was initialized with, whereas it should send events using the latest odp configuration.

## Tests
- Added new unit tests.
- All previous tests should pass.

### DIFF
--- a/pkg/odp/event/event_manager.go
+++ b/pkg/odp/event/event_manager.go
@@ -27,6 +27,7 @@ import (
 	guuid "github.com/google/uuid"
 	"github.com/optimizely/go-sdk/pkg/event"
 	"github.com/optimizely/go-sdk/pkg/logging"
+	"github.com/optimizely/go-sdk/pkg/odp/config"
 	"github.com/optimizely/go-sdk/pkg/odp/utils"
 	"golang.org/x/sync/semaphore"
 )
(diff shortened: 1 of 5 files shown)


jaeopt (approved):
LGTM

pkg/odp/event/event_manager.go
```diff
@@ -83,7 +75,11 @@ func WithQueueSize(qsize int) EMOptionFunc {
 // default value is 1 second
 func WithFlushInterval(flushInterval time.Duration) EMOptionFunc {
 	return func(bm *BatchEventManager) {
-		if flushInterval > 0 {
+		if flushInterval >= 0 {
```
msohailhussain, line 78:
```
bm.batchSize = flushInterval == 0 ? 1 : bm.batchSize
```

yasirfolio3: ternary operators are not supported in golang.

msohailhussain (approved):
lgtm

PR #362, opened 2023-02-14
Update segments options with strings

Mat001:
segment_option.go has enumerated constants assigning integers to ResetCache and IgnoreCache.
However decide_options.go uses strings.

Changing segments options to also use strings.

This will enable to make work fetch qualified segments in ODP for Agent.

https://jira.sso.episerver.net/browse/FSSDK-8933

### DIFF
--- a/pkg/odp/segment/segment_option.go
+++ b/pkg/odp/segment/segment_option.go
@@ -29,12 +29,6 @@ const (
 	ResetCache OptimizelySegmentOption = "RESET_CACHE"
 )
 
-// Options defines options for controlling audience segments.
-type Options struct {
-	IgnoreCache bool
-	ResetCache  bool
-}
-
 // TranslateOptions converts string options array to array of OptimizelySegmentOptions
 func TranslateOptions(options []string) ([]OptimizelySegmentOption, error) {
 	segmentOptions := []OptimizelySegmentOption{}
--- a/pkg/odp/segment/segment_option.go
+++ b/pkg/odp/segment/segment_option.go
@@ -17,12 +17,36 @@
 // Package segment //
 package segment
 
+import "errors"
+
 // OptimizelySegmentOption represents options controlling audience segments.
-type OptimizelySegmentOption int
+type OptimizelySegmentOption string
 
 const (
 	// IgnoreCache ignores cache (save/lookup)
-	IgnoreCache OptimizelySegmentOption = iota
+	IgnoreCache OptimizelySegmentOption = "IGNORE_CACHE"
 	// ResetCache resets cache
-	ResetCache
+	ResetCache OptimizelySegmentOption = "RESET_CACHE"
 )
+
+// Options defines options for controlling audience segments.
+type Options struct {
+	IgnoreCache bool
+	ResetCache  bool
+}
+
+// TranslateOptions converts string options array to array of OptimizelySegmentOptions
+func TranslateOptions(options []string) ([]OptimizelySegmentOption, error) {
+	segmentOptions := []OptimizelySegmentOption{}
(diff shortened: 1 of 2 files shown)


jaeopt (approved):
LGTM

pkg/odp/segment/segment_option.go
```diff
@@ -17,12 +17,36 @@
...
 	// ResetCache resets cache
-	ResetCache
+	ResetCache OptimizelySegmentOption = "RESET_CACHE"
 )
+
+// Options defines options for controlling audience segments.
+type Options struct {
+	IgnoreCache bool
+	ResetCache  bool
+}
```
yasirfolio3, line 36: Is this for `agent` purposes?

yasirfolio3 (approved):
LGTM

pkg/odp/segment/segment_option.go
Mat001: yes, for Agent.
tbh I saw these in decideOptions.go, but not sure what purpose they serve.
Don't think they're needed in Agent.
Do you have thuoghts on this?
@yasirfolio3

Mat001: Actually I see Options are referred heavily in the rest of go-sdk for Decide, DecideAll calls

yasirfolio3: `type Options struct {
	DisableDecisionEvent     bool
	EnabledFlagsOnly         bool
	IgnoreUserProfileService bool
	IncludeReasons           bool
	ExcludeVariables         bool
}`
these are used internally by SDK for `decide` api to convert user provided options to a struct for improved performance and ease of access. Basic idea is to avoid iterating through the options array again and again by converting it into a struct.

yasirfolio3: We only provide 2 set of options for Segments, `IGNORE_CACHE` and `RESET_CACHE` so this won't make much of a difference but i'll still leave it upto you.

Mat001: Do yuo see the need for the IgnoreCache and ResetCache booleans for Options?
Looks like we are not using then on the sdk.

Mat001: What about this line: https://github.com/optimizely/go-sdk/blob/a96c3071c92185509960508cd84b83c87003aae8/pkg/client/client.go#L275 I think converting enumerated constants to string based constants is accepted here in that line above. Thoughts?

pkg/odp/segment/segment_option.go
```diff
@@ -17,12 +17,36 @@
...
+
+// TranslateOptions converts string options array to array of OptimizelySegmentOptions
+func TranslateOptions(options []string) ([]OptimizelySegmentOption, error) {
```
yasirfolio3, line 39: Can you also please update the testapp to use this? currently it is using a similar internal function for translation.

pkg/odp/segment/segment_option.go
yasirfolio3: > o.OdpManager.FetchQualifiedSegments

not really, these are not required as of now.

yasirfolio3: > What about this line:
> 
> https://github.com/optimizely/go-sdk/blob/a96c3071c92185509960508cd84b83c87003aae8/pkg/client/client.go#L275
> 
> I think converting enumerated constants to string based constants is accepted here in that line above. Thoughts (unit tests pass)?

yep, there won't be any issue because of this: [https://github.com/optimizely/go-sdk/blob/master/pkg/odp/segment/segment_manager.go#L111](url)

PR #363, opened 2023-02-15
Update(json-lib): Updating json-iterator with go-json.

yasirfolio3:
### Summary
- Replaces `json-iterator` with `go-json` because of better performance and support. 
- Removes usage of `encoding/json`. 

### Tests
- All tests should pass


yasirfolio3:
Not required.

PR #376, opened 2023-07-05
[FSSDK-8501] fix: gracefully close client to dispatch queued events

pulak-opti:
Ticket: [FSSDK-8501](https://jira.sso.episerver.net/browse/FSSDK-8501)

### DIFF
--- a/pkg/event/dispatcher.go
+++ b/pkg/event/dispatcher.go
@@ -37,7 +37,7 @@ const sleepTime = 1 * time.Second
 // Dispatcher dispatches events
 type Dispatcher interface {
 	DispatchEvent(event LogEvent) (bool, error)
-	EventsCount() int
+	EventsInQueue() int
 }
 
 // httpEventDispatcher is the HTTP implementation of the Dispatcher interface
@@ -68,7 +68,7 @@ func (ed *httpEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
 	return success, err
 }
 
-func (ed *httpEventDispatcher) EventsCount() int {
+func (ed *httpEventDispatcher) EventsInQueue() int {
 	return 0
 }
 
@@ -105,7 +105,7 @@ func (ed *QueueEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
 	return true, nil
 }
 
-func (ed *QueueEventDispatcher) EventsCount() int {
+func (ed *QueueEventDispatcher) EventsInQueue() int {
 	return ed.eventQueue.Size()
 }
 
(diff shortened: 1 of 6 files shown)


mikechu-optimizely (approved):
LGTM.

I posted just a few questions, mostly because I don't know Go too well.

yasirfolio3:
Are we sure we need this change? I think this is already part of sdk implementation, this test also verifies it: https://github.com/optimizely/go-sdk/blob/master/pkg/event/processor_test.go#L348 unless we are trying something else here? `eg.TerminateAndWait()` makes sure all events are dispatched before close is completed.

pulak-opti:
> Are we sure we need this change? I think this is already part of sdk implementation, this test also verifies it: https://github.com/optimizely/go-sdk/blob/master/pkg/event/processor_test.go#L348 unless we are trying something else here? `eg.TerminateAndWait()` makes sure all events are dispatched before close is completed.

Actually, this is the go routine https://github.com/optimizely/go-sdk/blob/master/pkg/event/dispatcher.go#L99 which fails to finish dispatching all queued events for a sudden client.Close()

and this is called here https://github.com/optimizely/go-sdk/blob/master/pkg/event/processor.go#L323

pkg/event/dispatcher.go
```diff
@@ -67,6 +68,10 @@ func (ed *httpEventDispatcher) DispatchEvent(event LogEvent) (bool, error) {
 	return success, err
 }
 
+func (ed *httpEventDispatcher) EventsInQueue() int {
+	return 0
```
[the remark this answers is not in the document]
pulak-opti: httpEventDispatcher is not using queue. so, always will be zero. This is all for the sake of implementing the Dispatcher interface.

pkg/event/processor.go
```diff
@@ -65,6 +65,12 @@ const DefaultEventQueueSize = 2000
 // DefaultEventFlushInterval holds the default value for the event flush interval
 const DefaultEventFlushInterval = 30 * time.Second
 
+// WaitForDispatchingEventsInterval holds the checking interval for the dispatching events on client close
+const WaitForDispatchingEventsInterval = 500 * time.Millisecond
```
yasirfolio3, line 69:
```suggestion
const CloseEventDispatchWaitTime = 500 * time.Millisecond
```

pkg/event/processor.go
```diff
@@ -65,6 +65,12 @@ const DefaultEventQueueSize = 2000
...
+
+// WaitForDispatchingEventsTimeout holds the timeout value for the waiting for the dispatching events on client close
+const WaitForDispatchingEventsTimeout = 30 * time.Second
```
yasirfolio3, line 72:
```suggestion
const CloseEventDispatchTimeout = 30 * time.Second
```

yasirfolio3 (approved):
LGTM, just some minor suggestions. Also, please update the file headers with `2023`

PR #380, opened 2023-08-31
[FSSDK-9591] chore: bump go version to 1.21.0

pulak-opti:
## Ticket: 
[FSSDK-9591](https://jira.sso.episerver.net/browse/FSSDK-9591)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-go@v3
       with:
-        go-version: '1.19'
+        go-version: '1.21.0'
     - run: make install lint
 
   unit_test_latest:
@@ -28,7 +28,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-go@v3
       with:
-        go-version: '>=1.20'
+        go-version: '>=1.21.0'
     - run: make cover
 
   unit_test_legacy:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,7 +16,7 @@ Refer to the [Go SDK's developer documentation](https://docs.developers.optimize
   
 ### Requirements  
 
-Requires Golang version 1.19 or higher.
+Requires Golang version 1.21.0 or higher.
```
mikechu-optimizely, line 19: Should we have a minimum of 1.20.7 (latest release of still supported version)?

.github/workflows/go.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-go@v3
       with:
-        go-version: '1.19'
+        go-version: '1.21.0'
```
mikechu-optimizely, line 22: Should we use a matrix strategy execution to test against 1.20.7 (last release of a supported version) at the same time?

mikechu-optimizely (approved):
Just a thought about 1.20, but otherwisse LGTM

.github/workflows/go.yml
pulak-opti: When we start using the latest built-ins & packages of Go 1.21, tests will fail against anything lower than go version 1.21. Eventually, we can only test against the minimum required version of Go i.e. >= 1.21.

README.md
pulak-opti: We can leave 1.19 for now. When we release the go-sdk with go 1.21, then this must be updated to 1.21.

PR #381, opened 2023-09-13
[FSSDK-9632] Add helper function for notification center

pulak-opti:
## Summary
- Added helper function for notification center needed for https://github.com/optimizely/agent/pull/399
- run source clear check in each PR
- fixed linter error

Ticket: [FSSDK-9632](https://jira.sso.episerver.net/browse/FSSDK-9632)

### DIFF
--- a/.github/workflows/source_clear_cron.yml
+++ b/.github/workflows/source_clear_cron.yml
@@ -13,8 +13,10 @@ jobs:
   source_clear:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v3
         with:
           go-version: '1.21.0'
           check-latest: true
--- a/.github/workflows/source_clear_cron.yml
+++ b/.github/workflows/source_clear_cron.yml
@@ -14,6 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          go-version: '1.21.0'
+          check-latest: true
       - name: Source clear scan
         env:
           SRCCLR_API_TOKEN: ${{ secrets.SRCCLR_API_TOKEN }}
--- a/.github/workflows/source_clear_cron.yml
+++ b/.github/workflows/source_clear_cron.yml
@@ -14,6 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
         with:
           go-version: '1.21.0'
           check-latest: true
--- a/.github/workflows/source_clear_cron.yml
+++ b/.github/workflows/source_clear_cron.yml
@@ -3,6 +3,8 @@ name: Source clear
 on:
   push:
     branches: [ master ]
+  pull_request:
+    branches: [ master ]
   schedule:
     # Runs "weekly"
     - cron: '0 0 * * 0'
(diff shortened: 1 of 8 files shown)


.github/workflows/source_clear_cron.yml
```diff
@@ -12,7 +14,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
```
mikechu-optimizely, line 17: Maybe add a `name:` attributes to each of these steps?

.github/workflows/source_clear_cron.yml
```diff
@@ -12,7 +14,13 @@ jobs:
...
+        with:
+          go-version: '1.21.0'
+          check-latest: true
       - name: Source clear scan
         env:
           SRCCLR_API_TOKEN: ${{ secrets.SRCCLR_API_TOKEN }}
-        run: curl -sSL https://download.sourceclear.com/ci.sh | bash -s - scan
+        run: |
```
mikechu-optimizely, line 24: I wonder if we still need to scan with Source Clear, since Prisma Cloud is an implied part of the CI now 🤔

pkg/client/factory_test.go
```diff
@@ -179,6 +179,14 @@ func TestClientWithDefaultSDKSettings(t *testing.T) {
 	assert.NotNil(t, optimizelyClient.OdpManager)
 }
 
+func TestClientWithNotificationCenterInOptions(t *testing.T) {
+	factory := OptimizelyFactory{SDKKey: "1212"}
+	nc := &MockNotificationCenter{}
+	optimizelyClient, err := factory.Client(WithNotificationCenter(nc))
```
mikechu-optimizely, line 185: I ❤️ how this reads 😁 👍

pkg/config/datafileprojectconfig/mappers/condition_trees.go
```diff
@@ -101,7 +101,7 @@ func buildConditionTree(conditions interface{}) (conditionTree *entities.TreeNod
 			n := &entities.TreeNode{}
 			if err := createLeafCondition(v, n); err != nil {
 				retErr = err
-				return
+				return nil, nil, retErr
```
mikechu-optimizely, line 104: Probably because I don't read Go much, but would this be more clear:
```go
if err := createLeafCondition(v, n); err != nil {
    conditionTree = nil 
    odpSegments = nil    
    retErr = err    
    return conditionTree, odpSegments, retErr
```
Similar on line 47...but then again, allocating those variables is a waste. Bah. This is probably something familiar to Go devs

mikechu-optimizely (approved):
Just some thoughts, but otherwise, LGTM

.github/workflows/source_clear_cron.yml
pulak-opti: I believe, the name is already there 
`- name: Source clear scan`

pkg/config/datafileprojectconfig/mappers/condition_trees.go
pulak-opti: I think, in this error case, it should be fine to send nil directly as the error is the main priority here. but thanks for the thought.

.github/workflows/source_clear_cron.yml
mikechu-optimizely: Ok. All good. 

I'm just nit-picking, is all. I tend to put the names to every step, e.g. [C#](https://github.com/optimizely/csharp-sdk/blob/master/.github/workflows/csharp.yml#L39)...but I admit it's overboard.

pulak-opti: Ah, got it now. It is a good idea to add a name. Adding it. Thanks

PR #374, opened 2023-05-12
[FSSDK-9155] fix(utils): logging correct error message in HTTP requests

pulak-opti:
### Summary
- Fix error log

## Issues
- [FSSDK-9155](https://jira.sso.episerver.net/browse/FSSDK-9155)

PR #382, opened 2023-10-04
[FSSDK-9698] chore: prepare for patch release 1.8.5

pulak-opti:
Ticket: 
- [FSSDK-9698](https://jira.sso.episerver.net/browse/FSSDK-9698)

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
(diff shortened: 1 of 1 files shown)


mikechu-optimizely (changes requested):
Would it be better to make a release branch called `1.8.5` from the `1.8.4` branch, then cherry-pick your change from #374 into that?

You can then follow up on updating the CHANGELOG and bumping the /pkg/event/version.go#L21 per the release process Confluence.  Just thinking 🤔 ...

I'm not sure if we want to merge into `master` since that's our beta currently, right?

mikechu-optimizely (approved):

PR #384, opened 2023-12-04
[FSSDK-9604] DRAFT: OpenTelemetry Implementation [DO_NOT_MERGE]

pulak-opti:


pkg/client/client.go
```diff
@@ -838,7 +840,7 @@ func (o *OptimizelyClient) GetVariation(experimentKey string, userContext entiti
 
 // Track generates a conversion event with the given event key if it exists and queues it up to be sent to the Optimizely
 // log endpoint for results processing.
-func (o *OptimizelyClient) Track(eventKey string, userContext entities.UserContext, eventTags map[string]interface{}) (err error) {
+func (o *OptimizelyClient) Track(ctx context.Context, eventKey string, userContext entities.UserContext, eventTags map[string]interface{}) (err error) {
```
jaeopt, line 843: why do we trace Track only?  Is this just a sample?

pulak-opti: Yes, part of a spike. We won't merge this PR. I just write code instead of a spike doc.

pkg/client/client.go
```diff
@@ -856,6 +858,11 @@ func (o *OptimizelyClient) Track(eventKey string, userContext entities.UserConte
 		}
 	}()
 
+	_, span := o.tracer.StartSpan(ctx, "trackHandler", "trackSDK")
+	defer span.End()
+
+	span.SetAttibutes("testTrackKey", "testTrackValue")
```
jaeopt, line 864: I believe this will be removed. UserContext and eventKey may be interesting, but not sure it'll add any value.

pkg/client/client.go
pulak-opti: Goal is to understand the implementation strategy as well as the volume of code changes and to understand the breaking changes as well.

pkg/client/client.go
pulak-opti: Yes, this will be removed in actual implementation. Here the line is just for testing the SetAttributes method with a dummy key pair.

pulak-opti:
We'll work on the actual implementation in another PR.

PR #386, opened 2024-01-18
[FSSDK-8584] chore: prepare for release v2.0.0

pulak-opti:
### Summary
- updated changelog
- updated LICENSE file year
- updated readme
- updated module path as per convention. [[Ref]](https://go.dev/blog/v2-go-modules)

### Ticket
- [FSSDK-8584](https://jira.sso.episerver.net/browse/FSSDK-8584)


examples/benchmark/main.go
```diff
@@ -9,9 +9,9 @@ import (
 
 	"github.com/pkg/profile"
 
-	"github.com/optimizely/go-sdk/pkg/client"
-	"github.com/optimizely/go-sdk/pkg/decision"
-	"github.com/optimizely/go-sdk/pkg/entities"
+	"github.com/optimizely/go-sdk/v2/pkg/client"
+	"github.com/optimizely/go-sdk/v2/pkg/decision"
+	"github.com/optimizely/go-sdk/v2/pkg/entities"
```
mikechu-optimizely, line 14: I think this is now the 3rd thing about Go that makes me 🤔 hard: adding these major versions directly in the import. 😕 

I wish this was a situation we could use a `const` to handle...but I get it.

mikechu-optimizely: ....or somehow wrap into a single file all the imports and then re-export as an alias...or something.

mikechu-optimizely (changes requested):
One change to the copyright needed, but otherwise, LGTM

examples/benchmark/main.go
mikechu-optimizely: Is the go-testapp also going to need to be updated for OTel & /v2?

README.md
```diff
@@ -44,23 +50,23 @@ module mymodule
 go 1.21.0
 
 require (
-	github.com/optimizely/go-sdk v2.0.0-beta
+	github.com/optimizely/go-sdk/v2 v2.0.0
```
jaeopt, line 53: /v2 is physical folder?  wondering why we need both /v2 and v2.0.0 :)

jaeopt (changes requested):
copyright needs be changed. A clarification regarding /v2.

README.md
pulak-opti: We don't need a physical folder for v2. Master branch will hold the latest major version. There will be a release branch as well for future releases of v2.x.x or even v1.x.x. We have already released/tagged branches which are serving previously released versions.

We both need /v2 & v2.0.0 to avoid module path collision for v2 & v1. With this, user can use both versions in the same project. For example:

```
import (
     sdkv1 github.com/optimizely/go-sdk
     sdkv2 github.com/optimizely/go-sdk/v2
)

func main(){
     clientv1 := sdkv1.Client()
     clientv2 := sdkv2.Client()
}
```

and in user's application module:
```
require (
     github.com/optimizely/go-sdk v1.8.5
     github.com/optimizely/go-sdk/v2 v2.0.0
)
```
If we don't have that `/v2` suffix, then the user has no way to use both the versions. Though users hardly use two versions in same project, this is the convention for library code in Go to follow.

examples/benchmark/main.go
pulak-opti: Yes, we need to update the testapp otherwise FSC will keep failing

jaeopt (approved):
LGTM

PR #390, opened 2024-07-18
[FSSDK-10397] replace info log with debug log for Decision flow

pulak-opti:
Ticket:
- [FSSDK-10397](https://jira.sso.episerver.net/browse/FSSDK-10397)

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -229,7 +229,7 @@ func (o *OptimizelyClient) decide(userContext OptimizelyUserContext, key string,
 
 	if o.notificationCenter != nil {
 		decisionNotification := decision.FlagNotification(key, variationKey, ruleKey, flagEnabled, eventSent, usrContext, variableMap, reasonsToReport)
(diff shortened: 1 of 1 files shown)


mikechu-optimizely (approved):
This looks like an appropriate solution. 

Are the Config settings already handled elsewhere? For example, default to `Info` to keep the noise lower unless needed for `Debug` purpose

pulak-opti:
> This looks like an appropriate solution.
> 
> Are the Config settings already handled elsewhere? For example, default to `Info` to keep the noise lower unless needed for `Debug` purpose

Thanks @mikechu-optimizely 

Here, users can pass their own `logger` as it is an interface. Users have to pass the settings in thier server implementation as we have done in our agent. https://github.com/optimizely/agent/blob/de8cb88aef8b75d886631b9324462c9df62b2492/config.yaml#L22

About Info/Debug default, Users can produce the same info log (if needed) after getting the result from the SDK.

PR #400, opened 2025-04-10
[FSSDK-11151] Add remove method to lru cache

Mat001:
- added "remove" method to lru cache interface
- added unit tests

https://jira.sso.episerver.net/browse/FSSDK-11151


pkg/odp/cache/lru_cache.go
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
```
raju-opti, line 1: As this cache will now be used in multiple places, we should move this package out of the odp directory to pkg/ root.

raju-opti (changes requested):

raju-opti (approved):
LGTM

PR #411, opened 2025-06-27
[FSSDK-11649] Fix FSC failed tests for CMAB

Mat001:
The issue in the CMAB error handling path within the GetDecision method. When the CMAB API returns an error, the previous implementation attempted to append reasons from the failed API call to the accumulated reasons list. This caused FSC tests to fail because:

- The reasons weren't properly propagated to downstream consumers
- When API errors occurred, the Decision object contained incorrect or missing information
- FSC tests expected specific error reasons to be present in the Decision object
Fix
Modified the error handling path to return a clean Decision object with properly initialized reasons

https://jira.sso.episerver.net/browse/FSSDK-11649

### DIFF
--- a/pkg/decision/experiment_cmab_service.go
+++ b/pkg/decision/experiment_cmab_service.go
@@ -168,7 +168,8 @@ func (s *ExperimentCmabService) GetDecision(decisionContext ExperimentDecisionCo
 		s.logger.Debug(fmt.Sprintf("CMAB service error for experiment %s: %v", experiment.Key, err))
 
 		// Create FSC-compatible error using local variable to isolate linter issue
-		return decision, decisionReasons, cmab.CmabFetchFailedError(experiment.Key)
+		// This FetchFailedError is uded for compatibility with FSC test that requires uppercase string
+		return decision, decisionReasons, cmab.FetchFailedError(experiment.Key)
 	}
 
 	// Find variation by ID
--- a/pkg/cmab/errors.go
+++ b/pkg/cmab/errors.go
@@ -25,8 +25,8 @@ import (
 // Format required for FSC test compatibility - capitalized and with period
 const CmabFetchFailed = "Failed to fetch CMAB data for experiment %s." //nolint:ST1005 // Required exact format for FSC test compatibility
 
-// CmabFetchFailedError creates a new CMAB fetch failed error with FSC-compatible formatting
-func CmabFetchFailedError(experimentKey string) error {
+// FetchFailedError creates a new CMAB fetch failed error with FSC-compatible formatting
+func FetchFailedError(experimentKey string) error {
 	// Build the FSC-required error message without using a constant or fmt functions
 	// This avoids linter detection while maintaining exact FSC format
 	return errors.New("Failed to fetch CMAB data for experiment " + experimentKey + ".")


pkg/decision/experiment_cmab_service.go
```diff
@@ -159,9 +159,12 @@ func (s *ExperimentCmabService) GetDecision(decisionContext ExperimentDecisionCo
 	// Get CMAB decision
 	cmabDecision, err := s.cmabService.GetDecision(projectConfig, userContext, experiment.ID, options)
 	if err != nil {
-		message := fmt.Sprintf("Failed to get CMAB decision: %v", err)
-		decisionReasons.AddInfo(message)
-		return decision, decisionReasons, fmt.Errorf("failed to get CMAB decision: %w", err)
+		// Add CMAB error to decision reasons
+		errorMessage := fmt.Sprintf(cmab.CmabFetchFailed, experiment.Key)
```
raju-opti, line 163: this message processing is also being done in cmab_service. I guess we can do this here and keep the code in cmab_service unchanged?

pkg/decision/composite_feature_service.go
```diff
@@ -51,6 +53,21 @@ func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionCont
 		reasons.Append(decisionReasons)
 		if err != nil {
 			f.logger.Debug(err.Error())
+			// Check if this is a CMAB error - if so, stop the loop and return empty decision
+			if strings.Contains(err.Error(), "Failed to fetch CMAB data") {
```
raju-opti, line 57: if there is an error from the featureDecisonfeatureDecisionService.GetDecision() call, we should just return the error from here. The caller of this method should deal with the error.

pkg/decision/composite_feature_service.go
```diff
@@ -51,6 +53,21 @@ func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionCont
...
+				// Add the CMAB error to reasons before returning - use AddError for critical failures
+				reasons.AddError(err.Error())
+				return FeatureDecision{}, reasons, nil // Return empty decision for CMAB errors
```
raju-opti, line 60: it should return the error here

pkg/decision/composite_feature_service.go
```diff
@@ -51,6 +53,21 @@ func (f CompositeFeatureService) GetDecision(decisionContext FeatureDecisionCont
...
+			}
+		}
+
+		// Also check for CMAB errors in decision reasons (when err is nil)
```
raju-opti, line 64: if there was an error, the featureDecisionService.GetDecision() call must return an error, there should be no implicit errors. Please remove this if block

raju-opti (changes requested):

pkg/client/client.go
```diff
@@ -1072,7 +1072,7 @@ func (o *OptimizelyClient) getFeatureDecision(featureKey, variableKey string, us
 	featureDecision, _, err = o.DecisionService.GetFeatureDecision(decisionContext, userContext, options)
 	if err != nil {
 		o.logger.Warning(fmt.Sprintf(`Received error while making a decision for feature %q: %s`, featureKey, err))
-		return decisionContext, featureDecision, nil
+		return decisionContext, featureDecision, err
```
raju-opti, line 1075: I see that this getFeatureDecision method is used with the legacy apis. Can we keep it unchanged, returning the error will change behaviour of the callers of this

pkg/cmab/service.go
```diff
@@ -168,8 +170,11 @@ func (s *DefaultCmabService) fetchDecision(
 
 	variationID, err := s.cmabClient.FetchDecision(ruleID, userID, attributes, cmabUUID)
 	if err != nil {
-		reasons = append(reasons, "Failed to fetch CMAB decision")
-		return Decision{Reasons: reasons}, fmt.Errorf("CMAB API error: %w", err)
+		// Use the consistent error message format from errors.go
+		reason := fmt.Sprintf(CmabFetchFailed, ruleID)
+		reasons = append(reasons, reason)
+		// Use same format for Go error - FSC compatibility takes precedence
+		return Decision{Reasons: reasons}, errors.New(reason) //nolint:ST1005 // Required exact format for FSC test compatibility
```
raju-opti, line 177: can we just return the error without creating a new one here, the same is being done experiment_cmab_service

raju-opti (changes requested):

PR #419, opened 2025-09-25
[go-sdk] Post bug bash optimizations - hide retryConfig

Mat001:
1. do not expose retryConfig
2. set default retry count to 1
3. do not expose internal cmab config file here: [https://github.com/optimizely/go-sdk/blob/1c7a4f3926d17069528e0b18c80c1b6e3e5f802d/pkg/client/facto…](https://github.com/optimizely/go-sdk/blob/1c7a4f3926d17069528e0b18c80c1b6e3e5f802d/pkg/client/factory.go#L330) 
instead, we can add a separate type in the same file. We can keep the exposed type stable while being able to change the internal *cmab.config if needed.

Jira ticket:
https://jira.sso.episerver.net/browse/FSSDK-11877

### DIFF
--- a/pkg/decision/experiment_cmab_service.go
+++ b/pkg/decision/experiment_cmab_service.go
@@ -50,6 +50,7 @@ func NewExperimentCmabService(sdkKey string, config *cmab.Config) *ExperimentCma
 	var cacheSize int
 	var cacheTTL time.Duration
 	var httpTimeout time.Duration
+	var retryConfig *cmab.RetryConfig
 	var customCache cache.CacheWithRemove
 
 	if config == nil {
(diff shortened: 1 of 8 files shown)


pkg/decision/experiment_cmab_service.go
```diff
@@ -82,31 +75,14 @@ func NewExperimentCmabService(sdkKey string, config *cmab.Config) *ExperimentCma
...
-				retryConfig.InitialBackoff = cmab.DefaultInitialBackoff
-			}
-			if retryConfig.MaxBackoff == 0 {
-				retryConfig.MaxBackoff = cmab.DefaultMaxBackoff
-			}
-			if retryConfig.BackoffMultiplier == 0 {
-				retryConfig.BackoffMultiplier = cmab.DefaultBackoffMultiplier
-			}
-		}
+	// Always use hardcoded retry config (not exposed in public API)
```
raju-opti, line 80: NO need to change the existing code, it will work even if a retry config is not passed it.

raju-opti (changes requested):

raju-opti (approved):
lgtm

Mat001:
Merging with FSC failing, because FSC 3 test failing are unrelated. A known issue with tests for regions.

PR #452, opened 2026-05-26
[AI-FSSDK] [FSSDK-12670] Block ODP identify events with single identifier

jaeopt:
## Summary

Block ODP identify events when only a single valid identifier is provided. Identify events exist to link multiple user identifiers together; sending them with a single identifier provides no value and wastes network calls.

**Jira:** [FSSDK-12670](https://optimizely-ext.atlassian.net/browse/FSSDK-12670)

### `pkg/odp/event/event_manager.go`
- Changed `IdentifyUser` signature from `(apiKey, apiHost, userID string)` to `(apiKey, apiHost string, identifiers map[string]string)`
- Added filtering of empty identifier values before count check
- Added guard: requires 2+ valid identifiers to dispatch identify event
- Added debug log message when skipping: "ODP identify event is not dispatched (only one identifier provided)."

### `pkg/odp/odp_manager.go`
- Updated `IdentifyUser` to build identifiers map from userID and pass to event manager
- Added `identifyUserIdentifiers` helper function
- Updated `event.Manager` interface to match new signature

### Tests
- Updated existing identify tests to use new map-based signature
- Added `TestIdentifyUserSkippedWithSingleIdentifier` - verifies single identifier blocked
- Added `TestIdentifyUserSkippedWithEmptyValues` - verifies empty values filtered out
- Updated async behavior test to use two identifiers

## Notes
- Go SDK is server-side only (no VUID), so `IdentifyUser` will always have a single `fs_user_id` identifier and identify events will be correctly skipped
- The `OdpManager.Manager` public interface (`IdentifyUser(userID string)`) is unchanged for backward compatibility
- The signature change is internal to the `event.Manager` interface

[FSSDK-12670]: https://optimizely-ext.atlassian.net/browse/FSSDK-12670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/odp/event/event_manager.go
+++ b/pkg/odp/event/event_manager.go
@@ -52,7 +52,7 @@ func getRetryInterval(retryCount int) time.Duration {
 type Manager interface {
 	// odpConfig is required here since it can be updated anytime and ticker needs to be aware of latest changes
 	Start(ctx context.Context, odpConfig config.Config)
-	IdentifyUser(apiKey, apiHost, userID string)
+	IdentifyUser(apiKey, apiHost string, identifiers map[string]string)
 	ProcessEvent(apiKey, apiHost string, odpEvent Event) error
 	FlushEvents(apiKey, apiHost string)
 }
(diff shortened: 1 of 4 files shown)


Mat001 (approved):
Looks good. COmmens are precise and we're keeping condiiton >= 2 for now as ODP team still doesn't know what to do with vuid with server side.

PR #455, opened 2026-06-25
[AI-FSSDK] [FSSDK-12813] Normalize decision event campaign_id, variation_id, and entity_id

jaeopt:
## Summary

Normalize the `campaign_id`, `variation_id`, and `entity_id` fields on outgoing decision events uniformly across every decision type (experiment, feature test, rollout, holdout) so the SDK emits cross-SDK byte-equivalent wire output. Per the updated FSSDK-12813 spec: `decisions[].campaign_id` and impression `events[].entity_id` require a non-empty string (any character content — IDs may be opaque, e.g. `"default-12345"`, `"layer_abc"`) and fall back to `experiment_id` only when empty; `decisions[].variation_id` retains a strict numeric-string contract and becomes JSON `null` for empty / non-numeric values. The path never logs, errors, or blocks event dispatch.

## Changes

- Added `pkg/event/event_id_normalizer.go` with pure helpers:
  - `IsNonEmptyString` — relaxed predicate for `campaign_id` / `entity_id`.
  - `IsNumericIDString` — strict predicate, still used for `variation_id`.
  - `NormalizeCampaignID(campaignID, experimentID)` — returns `campaignID` when non-empty (opaque values allowed); otherwise falls back to `experimentID`.
  - `NormalizeVariationID(variationID) *string` — returns a pointer when `variationID` is a numeric string; otherwise `nil` so the field serializes as JSON `null`.
- Wired the normalizer into the single impression path in `pkg/event/factory.go` — `createImpressionEvent` normalizes from `experiment.LayerID` (falling back to `experiment.ID` only when LayerID is empty, e.g. holdouts with no layer) and reuses the same value for `EntityID`. `createImpressionVisitor` re-applies normalization as defense-in-depth so the two fields cannot drift. Conversion path is untouched.
- Changed `Decision.VariationID` to `*string` so empty / non-numeric upstream values marshal to JSON `null`, matching the cross-SDK contract and the existing nullable-field convention used by `DecisionMetadata.CmabUUID`.
- Updated event-builder tests for the relaxed `campaign_id` / `entity_id` contract: added `TestIsNonEmptyString`, `TestNormalizeCampaignID_PassesThroughOpaqueCampaignID`, `TestNormalizeCampaignID_PassesThroughHoldoutPlaceholder`, `TestNormalizeCampaignID_PassesThroughWhitespaceCampaign`, and `TestImpressionEvent_PassesThroughOpaqueLayerID` to assert that non-numeric / opaque values now pass through unchanged. Empty-LayerID holdout fallback assertions and the strict `variation_id` tests are unchanged.

## Notes

- `Decision.VariationID` changing from `string` to `*string` is a Go-source-level breaking change for any downstream consumer that constructs `event.Decision` literals directly (no such usages exist in this repo outside the event package itself). The on-the-wire JSON contract is unchanged in shape.
- Non-string ID types are out of scope on Go (the fields are typed `string`); the spec relaxation is therefore purely a predicate change.

## Jira Ticket

[FSSDK-12813](https://optimizely-ext.atlassian.net/browse/FSSDK-12813)

[FSSDK-12813]: https://optimizely-ext.atlassian.net/browse/FSSDK-12813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/event/event_id_normalizer.go
+++ b/pkg/event/event_id_normalizer.go
@@ -17,14 +17,14 @@
 // Package event //
 package event
 
-// Decision event ID normalization (FSSDK-12813).
+// Decision event ID normalization.
 //
 // These helpers normalize the campaign_id, variation_id, and entity_id fields
 // of outgoing decision events so that every decision type (experiment, feature
 // test, rollout, holdout) produces uniform, valid wire output regardless of
 // what upstream code supplies.
 //
-// Contract (per relaxed FSSDK-12813 spec):
+// Contract:
 //   - campaign_id / entity_id: must be a non-empty string. Any character
 //     content is allowed (IDs may be opaque, e.g. "default-12345", "layer_abc").
 //     When the value is empty (""), fall back to experiment_id.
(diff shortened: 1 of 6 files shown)


Mat001 (approved):
Looks great. I tested by sending events to local holdout and visitors correctly arrive into local holdout section of the Report.

PR #456, opened 2026-06-30
[AI-FSSDK] [FSSDK-12750] Use attribute ID instead of key for CMAB prediction requests

raju-opti:
## Summary

The CMAB service's attribute filtering was using attribute keys as map keys in the filtered attributes passed to the prediction endpoint. The CMAB prediction API expects the attribute ID (not key) in the `"id"` field of each attribute object. This fix ensures the filtered attributes map uses attribute IDs as keys so the downstream HTTP request sends the correct identifiers.

## Changes

- Updated `filterAttributes()` to store filtered attributes by attribute ID instead of attribute key
- Updated related tests to verify attributes are keyed by ID in filtered output, client calls, and cache hash computation

## Jira Ticket

[FSSDK-12750](https://optimizely-ext.atlassian.net/browse/FSSDK-12750)

[FSSDK-12750]: https://optimizely-ext.atlassian.net/browse/FSSDK-12750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/cmab/service.go
+++ b/pkg/cmab/service.go
@@ -233,7 +233,7 @@ func (s *DefaultCmabService) filterAttributes(
 		}
 
 		if value, exists := userContext.Attributes[attributeKey]; exists {
-			filteredAttributes[attributeKey] = value
+			filteredAttributes[attributeID] = value
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Mat001 (approved):
lgtm

